PR #18, opened 2016-06-02
Python 3.6.0a1 recipe

jjhelmus:


jjhelmus:
This should be merged into a new branch named `python_3.6` or similar but I don't think this is possible with the GitHub UI.

jjhelmus:
The Linux and OS X builds are failing because the test phase is trying to to install pip and a Python 3.6 pip does not exist.  Adding `add_pip_as_python_dependency: false` to the .condarc file should fix this.  

Not sure what is going on with the Windows builds.

jakirkham:
Ok, so, you're seeing this too. Do you have any idea why `activate root` doesn't work on Windows? I would expect this to just work, but maybe I just don't get something about `conda` on Windows.

cc @msarahan

jakirkham:
Maybe we just `deactivate` as this [thread]( https://groups.google.com/a/continuum.io/forum/#!topic/anaconda/iZoXAa8ZImU ) suggests?

msarahan:
Nice, @jjhelmus!

For activating root, you can just:

```activate``` with no args.

The activation behavior will be normalized with conda 4.1 - both activate root and activate with no args should work at that point.

jakirkham:
> For activating root, you can just:

> ```activate``` with no args.

Awesome. Thanks.

> The activation behavior will be normalized with conda 4.1 - both activate root and activate with no args should work at that point.

😌

jakirkham:
I don't think it is running the activation scripts on Windows. Did we configure them wrong? Is there something we are missing?

jjhelmus:
I'm still at PyCon and do not have access to my Windows box at the moment so testing that build is a bit difficult.

jjhelmus:
Also conda-forge/staged-recipes#705 is probably a better test of this since the Windows build of toolz actually works where as Python 3.6 is currently broken for a unrelated reason.

jakirkham:
Agreed. Just trying to seeing advice from those that know Windows better. 😄

jjhelmus:
All green 🎉  Until we decide how to deal with pre-releases this should not be merge.

jjhelmus:
Closing in favor of #43

PR #19, opened 2016-06-03
build python with external libraries

msarahan:
As usual, cramming too much into one PR.

The intent here is to make Python not pull in its own dependencies and build them statically, but rather provide them with our versions.

This PR addresses only Windows, but seems to work on Windows.

I have unpinned a few things to make it work.  Side note: @183amir, your vc package was a lifesaver here.  I could not get features to work correctly without it.

CC @mingwandroid

### DIFF
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -7,15 +7,17 @@ REM Compile python, extensions and external libraries
 if "%ARCH%"=="64" (
    set PLATFORM=x64
    set VC_PATH=x64
-   set BUILD_PATH=amd64
+   set BUILD_PATH=amd64-pgo
 ) else (
    set PLATFORM=Win32
    set VC_PATH=x86
-   set BUILD_PATH=win32
+   set BUILD_PATH=win32-pgo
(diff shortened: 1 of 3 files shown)


jakirkham:
Glad to hear that the `vc` package is coming in handy.

jjhelmus:
This seems to depend on a tk 8.6 package which does not yet exist.  I believe conda-forge/tk-feedstock#1 needs to be merged and a package created before this can proceed.

jakirkham:
So we have 8.6.x. Should we resolve merge conflicts and try again?

jakirkham:
Came across this in a recent search. Curious if this is still needed (and should be refreshed) or if not (and should be closed). Thoughts?

jjhelmus:
This still would be nice to have.  Among other benefits, it would allow OpenSSL to be upgraded without needing to rebuild Python.  That said, the changes needed to the build files are relatively involved and would likely need to be updated for each new Python version, the benefits may not be worth the effort.

PR #11, opened 2016-05-20
WIP: Fix windows 64-bit build

jjhelmus:

PR #20, opened 2016-06-14
Rerender the feedstock

jjhelmus:
closes #11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ language: objective-c
 env:
   matrix:
     
-    - CONDA_PY=351
+    - CONDA_PY=35
   global:
     # The BINSTAR_TOKEN secure variable. This is defined canonically in conda-forge.yml.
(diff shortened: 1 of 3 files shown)


jjhelmus:
See #11 for a work on debugging this issue and an explanation of the fix.

jjhelmus:
I don't think AppVeyor actually needs quotes, apparently conda 4.1.0 broke something in the test phase that was working in 4.0.8

jakirkham:
Looks like you might be right. Going to try to reproduce it in another case just to be sure.

jjhelmus:
Going to leave this be for a bit until the conda issue gets worked out.

jjhelmus:
So all the builds are good, is this worth merging so that the win-64 builds goes out or should we wait until a conda 4.1.1 release is made and the order of adding and list channels is no longer needed?

msarahan:
Thanks for figuring this mess out.

PR #24, opened 2016-06-25
OS X framework build

jjhelmus:
On OS X Python is built as a framework.  Patches and ./configure setting taken
from MacPorts project
[skip appveyor]

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -6,8 +6,8 @@ if [ `uname` == Darwin ]; then
     export CFLAGS="-I$PREFIX/include $CFLAGS"
     export LDFLAGS="-L$PREFIX/lib -headerpad_max_install_names $LDFLAGS"
     sed -i -e "s/@OSX_ARCH@/$ARCH/g" Lib/distutils/unixccompiler.py
-    ./configure --enable-shared --enable-ipv6 --with-ensurepip=no \
-        --prefix=$PREFIX
+    ./configure --enable-ipv6 --with-ensurepip=no \
+        --prefix=$PREFIX --enable-framework=$PREFIX
 fi
 if [ `uname` == Linux ]; then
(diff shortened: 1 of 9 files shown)


jjhelmus:
If anyone wants to try out the package I've uploaded it to [my (jjhelmus) Anaconda cloud channel](https://anaconda.org/jjhelmus/python/files) as `osx-64/python-3.5.1-1.tar.bz2`.

jakirkham:
One thought to avoid compatibility issues, assuming we want to go down this road (and I'm still not sure that I and/or we want to), is to move all of the contents of the framework into the standard locations and replace everything in the framework with links. That way, if the framework is not present, it does not necessarily cause issues. Note that the framework may not necessarily be present as people switch back and forth between `defaults` versions of the the package and `conda-forge` versions.

If we did that, it actually might also be possible to think of the framework as a separate package that is installed on top of a working Python distribution to provide this added flexibility. This would also help avoid some of the aforementioned issues.

Thoughts on this point?

jjhelmus:
@jakirkham I agree, there are going to be a compatibility issues with a framework build.  It may be possible to work around them but we would need to maintain these changes and throughly test them for compatibility with the `defaults` version of Python and other conda packages.  I believe these tasks would require  a significant amount of time and personally do not think they are worth the effort.

jakirkham:
Well, we can always investigate the wrapper script idea, which does sound considerably simpler. Though I fully agree that these are both going to take a bit of work to properly vet. In fact, as we are concerned about stability, I would argue whatever we change we propose not even be part of the normal `python` package, but be an add-on package that users must chose to install. This way users know they are getting into weird territory.

jjhelmus:
Closing this PR.  

I have no plans to pursue a OS X framework build of Python.  If anyone else is interesting in working on this please take over.

PR #25, opened 2016-06-27
Various OS X related improvements

jjhelmus:
Create a new OS X build which incorporated a number of minor bug fixed and recipe improvements.

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -4,7 +4,7 @@ python ${RECIPE_DIR}/brand_python.py
 
 if [ `uname` == Darwin ]; then
     export CFLAGS="-I$PREFIX/include $CFLAGS"
-    export LDFLAGS="-L$PREFIX/lib -headerpad_max_install_names $LDFLAGS"
+    export LDFLAGS="-Wl,-rpath,$PREFIX/lib -L$PREFIX/lib -headerpad_max_install_names $LDFLAGS"
     sed -i -e "s/@OSX_ARCH@/$ARCH/g" Lib/distutils/unixccompiler.py
     ./configure --enable-shared --enable-ipv6 --with-ensurepip=no \
         --prefix=$PREFIX
@@ -22,17 +22,3 @@ make
 make install
 ln -s $PREFIX/bin/python3.5 $PREFIX/bin/python
 ln -s $PREFIX/bin/pydoc3.5 $PREFIX/bin/pydoc
-
-if [ `uname` == Darwin ]; then
-    # Some extension modules are renamed when importing fails do to a missing
-    # or incorrect shared library. These libraries are available through conda,
-    # but their locations must first be corrected in the post-build step using
-    # install_name_tool. Un-do this renaming so that the modules can be
-    # imported after the library locations are corrected.
-    pushd $PREFIX/lib/python3.5/lib-dynload
-    for filename in *_failed.so; do
-        [ -f "$filename" ] || continue;
-        mv $filename ${filename//_failed/};
-    done
-    popd
-fi
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -6,8 +6,13 @@ if [ `uname` == Darwin ]; then
     export CFLAGS="-I$PREFIX/include $CFLAGS"
     export LDFLAGS="-Wl,-rpath,$PREFIX/lib -L$PREFIX/lib -headerpad_max_install_names $LDFLAGS"
     sed -i -e "s/@OSX_ARCH@/$ARCH/g" Lib/distutils/unixccompiler.py
-    ./configure --enable-shared --enable-ipv6 --with-ensurepip=no \
-        --prefix=$PREFIX
+    ./configure \
+        --enable-ipv6 \
+        --enable-shared \
+        --prefix=$PREFIX \
+        --with-ensurepip=no \
+        --with-tcltk-includes="-I$PREFIX/include" \
+        --with-tcltk-libs="-L$PREFIX/lib -ltcl8.5 -ltk8.5"
 fi
 if [ `uname` == Linux ]; then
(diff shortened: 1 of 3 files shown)


jjhelmus:
Thanks to @jakirkham for a number of this fixes

jjhelmus:
Because the libraries in the `openssl` and `readline` packages from `defaults` do not use `@rpath` in their names, the fix in 7674688 requires a bit more work.  Luckily the `conda-forge` versions of these packages do use `@rpath` in the library names.  

One solution, if the `defaults` packages are used, is to modify the library names in the build environment as proposed by @pkgw in ContinuumIO/anaconda-issues/issues/598.  In this case the following lines would need to be added to the OS X section of the `build.sh` file

```
mv $PREFIX/lib/libreadline.6.2.dylib $PREFIX/lib/tmp
cp $PREFIX/lib/tmp $PREFIX/lib/libreadline.6.2.dylib
rm -f $PREFIX/lib/tmp
install_name_tool -id @rpath/libreadline.6.2.dylib $PREFIX/lib/libreadline.6.2.dylib

mv $PREFIX/lib/libssl.1.0.0.dylib $PREFIX/lib/tmp
cp $PREFIX/lib/tmp $PREFIX/lib/libssl.1.0.0.dylib
rm -f $PREFIX/lib/tmp
install_name_tool -id @rpath/libssl.1.0.0.dylib $PREFIX/lib/libssl.1.0.0.dylib
mv $PREFIX/lib/libcrypto.1.0.0.dylib $PREFIX/lib/tmp
cp $PREFIX/lib/tmp $PREFIX/lib/libcrypto.1.0.0.dylib
rm -f $PREFIX/lib/tmp
install_name_tool -id @rpath/libcrypto.1.0.0.dylib $PREFIX/lib/libcrypto.1.0.0.dylib
```

recipe/meta.yaml
```diff
@@ -5,10 +5,14 @@ package:
...
-  md5: be78e48cdfc1a7ad90efff146dce6cfe
+  sha256: 687e067d9f391da645423c7eda8205bae9d35edc0c76ef5218dcbe4cc770d0d7
+  patches:
+      - omit-local-site-packages.patch  # [osx]
 
 build:
-  number: 0
+  number: 0  # [win]
+  number: 0  # [linux]
+  number: 1  # [osx]
```
jakirkham: So, I know we are only changing OS X here, but I'm ok with using one build number to keep life simple.

jakirkham:
Of course. Sorry for leaving them sprawled out everywhere. Thanks for pulling them together.

recipe/meta.yaml
jjhelmus: I have a feeling that these will go out of sync pretty quickly.  I though might as well separate them now, but I'm okay either way

recipe/build.sh
```diff
@@ -4,10 +4,15 @@ python ${RECIPE_DIR}/brand_python.py
...
     sed -i -e "s/@OSX_ARCH@/$ARCH/g" Lib/distutils/unixccompiler.py
-    ./configure --enable-shared --enable-ipv6 --with-ensurepip=no \
-        --prefix=$PREFIX
+    ./configure \
+        --enable-ipv6 \
+        --enable-shared \
+        --prefix=$PREFIX \
+        --with-ensurepip=no \
+        --with-tcltk-includes="-I$PREFIX/include" \
+        --with-tcltk-libs="-L$PREFIX/lib -ltcl8.5 -ltk8.5"
```
jakirkham: Should we try `--without-gcc`?

jjhelmus: At least on my machine then the readline, ssl and _hashlib modules fail to import.  I don't think clang's linker takes the `-Wl,-rpath` argument. Is there a compelling reason not to use gcc?

jakirkham: It's fine. We can explore this in another patch.

Not that this is very important at all, but was just trying to get this to use `clang` as well. Though it doesn't use C++, so it shouldn't really matter which compiler it uses.

jakirkham: Is it possible that you have an older version of `conda-build`? There was a bug where it chopped stuff out of your environment variables not that long ago. It should be fixed in more recent versions.

jjhelmus: Nope, using the latest versions of conda (4.1.4) and conda-build (1.21.2).  I say we keep this one for a different patch.

jjhelmus: Axe that, I forgot to pull in the `conda-forge` version of the readline and openssl packages which was causing the issue.  `--without-gcc` does work if you disable the lines in `run_test.py` that check for gcc and g++. 

So we could add this argument but I'm worried that changing the sysconfig settings might break downstream packages which uses these for their builds.  Thoughts?

jakirkham: So, the other reason I was asking for this flag is I have found things that do link to the system `gcc`'s libraries often don't work on 10.11 (maybe the libraries are removed? will need to check when I have 10.11 system to try with). Again I don't see why it should cause issues for us with `python` given it is just C, but it is something to be mindful of.

jjhelmus: I'll keep it in mine.  If it crops up we can revisit.

jjhelmus:
Lets see if the new commit passes but I think this is good to go after that.  Any other comments @jakirkham?

jakirkham:
Nope. LGTM.

pkgw:
I'm not fully following this discussion, but for what it's worth, here's a list of the libraries that I've found in Anaconda that have issues with their install names. Not guaranteed to be complete.

- crypto
- curl
- freetype
- png16
- readline
- ssl
- xml2
- xslt

It's been a little frustrating since the Continuum folks seem to be very unresponsive when it comes to fixing packaging bugs like ContinuumIO/anaconda-issues#598.

PR #28, opened 2016-07-01
Python 1.0.1

jjhelmus:
[skip appveyor]


jjhelmus:
This should not be merged into master rather into a "version" branch.  I believe this must be done during the merge rather than in the pull request.

jakirkham:
Woah! What?

Should we maybe not put this into `master` if this is happening?

jakirkham:
Sorry, I think we posted at the same time. So, we can add a new branch for `1.x` that you can PR against. Would you like to do this?

jjhelmus:
That will work, my plan was to push this branch to upstream/1.0 if the CIs pass

jakirkham:
Alright, let me create a branch and then we can move the PR over to it.

jakirkham:
Alright, there is a `1.0` branch that is at the same point as `master`. Could we please PR against that instead?

jjhelmus:
Also in case anyone is interests the python-1.0.1 package in `defaults` does not properly set `sys.path` which means most of the "standard library" is not available... although I don't really think this will ever effect anyone.

jjhelmus:
I'll look into the OS X issue and submit a new PR against the `1.0` branch.

PR #29, opened 2016-07-01
Python version 1.0.1

jjhelmus:


jjhelmus:
Additionally, the core of the filesystem uses Case Insensitive Unicode Strings, which led Linus Torvalds to state "HFS+ is probably the worst file-system ever."

[Wikipedia: HDF Plus](https://en.wikipedia.org/wiki/HFS_Plus)

jjhelmus:
Oops, forgot to skip AppVeyor, but Travis and Circle are passing. This should be good to go if anyone wants to review.

jjhelmus:
Closing, the Python 1.X line is being built in the `historic` branch, for example #44.

PR #31, opened 2016-07-01
Python 1.2

jjhelmus:
[skip appveyor]


jjhelmus:
Should a libcrypt package be provided or is the library common enough to be found on most linux systems?  I believe this is the only library that may need to be packaged. The complete linking for the python binary is:

```
$ ldd ~/anaconda/envs/py12/bin/python 
	linux-vdso.so.1 =>  (0x00007ffdb079c000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fbb4ecc4000)
	libcrypt.so.1 => /lib/x86_64-linux-gnu/libcrypt.so.1 (0x00007fbb4ea8b000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fbb4e887000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fbb4e4c2000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fbb4efca000)
```

jakirkham:
Based on this [info]( http://www.linuxquestions.org/questions/linux-software-2/what-is-lib-libcrypt-766840/ ) it sounds like `libcrypt` is part of `glibc`. I don't think we should be packaging it then. Also, I think we can safely assume the user has this on the system.

Just out of curiosity, are you expect lots of use for these old Pythons or are you wanting to build them more as curiosities for people to explore?

jakirkham:
Also, if we are changing the `skip` clause, we need to run `conda smithy rerender` to adjust this image. I think it will just come down to deleting the `appveyor.yml` file though.

jjhelmus:
> Just out of curiosity, are you expect lots of use for these old Pythons or are you wanting to build them more as curiosities for people to explore?

Curiosity more than anything, having a three day holiday weekend also helps. I don't expect these to be used much besides toy examples although I have a fun talk in mind that I might use them for.  Seeing how the build system has changed over the year has been interesting.

jjhelmus:
I think these Python 1.x builds are ready for review and merging.  Python 1.1 is a pain to compile on 64-bit system as it assumes that the va_list structure in `stdarg.h` is not an array, which it is on most system.  I don't expect this to be a big loss.

jjhelmus:
closed in favor of #48 which merged these changed into the `historic` branch.

PR #43, opened 2016-07-12
Python 3.6.0a3

jjhelmus:


jjhelmus:
Closes in favor of #45 which merges into prereleases

PR #51, opened 2016-07-23
Various Python 3.5 fixes

jjhelmus:
Makes python 3.5 recipe similar to recipe in anaconda-recipes


ocefpaf:
@jjhelmus I got my hands on a Win64 machine for the weekend. Let me know if you need help testing things.

jjhelmus:
Apparently the patches in `anaconda-recipes` are not always valid.  Will need to make my own.

ocefpaf:
> Apparently the patches in `anaconda-recipes` are not always valid.  Will need to make my own.

I guess we should start packaging (and using) [`quilt`](http://savannah.nongnu.org/projects/quilt) to manage these.

jjhelmus:
The patch problem is not really the fault of `anaconda-recipes` as they are applying an additional patch to `site.py` which when these two patches are applied in the correct order resolves fine.  We don't want the `site.patch` patch from anaconda-recipes so regeneration of the other patch was needed.

recipe/meta.yaml
```diff
@@ -8,14 +8,17 @@ source:
   sha256: 1524b840e42cf3b909e8f8df67c1724012c7dc7f9d076d4feef2d3eff031e8a0
   patches:
       - omit-local-site-packages.patch  # [osx]
+      - osx64-dist.patch                # [osx and x86_64]
```
jakirkham: Though I don't think OS X 32-bit has been supported for a bit. I don't have any problem with adding `x86_64`.

jjhelmus: I agree, for use in `conda-forge` the `x86_64` tag is not needed but adding it makes the recipe match better with the one in `anaconda-recipes`.

jjhelmus:
Any other comments on this issue?  I was planning on merging this later today unless there is a show stopper.  

I'll submit a similar update to Python 3.4 shortly.

recipe/meta.yaml
```diff
@@ -8,14 +8,17 @@ source:
...
 build:
-  number: 1
+  number: 2
   # Windows has issues updating python if conda is using files itself.
   # Copy rather than link.
   no_link:
-    - DLLs/_ctypes.py   # [win]
-    - bin/python
+    - DLLs/_ctypes.pyd  # [win]
+    - bin/python3.5     # [unix]
```
jakirkham: What's going on with this?

jjhelmus: Both of those lines were taken from the [python 3.5 recipe on anaconda-recipes](https://github.com/ContinuumIO/anaconda-recipes/blob/master/python-3.5/meta.yaml)

bin/python is a symbolic link to bin/python3.5 so it makes more sense to not link the true file as opposed to the symbolic link. 

DLLs/_ctypes.py does not exist and it should be _ctypes.pyd, just a typo in the original recipe that was never caught.  I believe it must be because of the comment above in the souce.

jjhelmus:
Is this good to go?

jakirkham:
Fine by me. Should we have someone look over the Windows stuff? I'm still fumbling in that area so might not be the best reviewer for that.

msarahan:
Thanks @jjhelmus

PR #67, opened 2016-09-06
Enable branding environment variable inheritance.

bjlittle:
Enable the recipe to inherit the `python_branding` environment variable for customisation within the  `brand_python.py` script.

https://github.com/conda-forge/python-feedstock/blob/master/recipe/brand_python.py#L36

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -13,7 +13,7 @@ source:
       - win-library_bin.patch           # [win]
 
 build:
-  number: 2
+  number: 3
   # Windows has issues updating python if conda is using files itself.
   # Copy rather than link.
   no_link:
@@ -21,6 +21,8 @@ build:
     - bin/python3.5     # [unix]
   track_features:
     - vc14              # [win]
+  script_env:
+    - python_branding
 
 requirements:
   build:


bjlittle:
@jjhelmus or @msarahan fancy considering this PR? Thanks

pelson:
:+1: - LGTM (but I'm not going to merge due to conflict of interest). @bjlittle - did you also want this for the 2.7 branch?

bjlittle:
@pelson Thanks, I'll also target `2.7` and `3.4` in separate PRs

jakirkham:
All of these seem reasonable to me. Not entirely sure how they would work without this though. Should we wait for @jjhelmus who authored these recipes to double check we aren't missing something obvious or are you happy enough to have them merged now?

bjlittle:
> Not entirely sure how they would work without this though.

@jakirkham Yeah, this is the only way that I can see for the python branding to work. I hope that I'm not missing something obvious (which I'd love to know, if that's the case!)

Thanks

bjlittle:
Ping @jakirkham @msarahan @jjhelmus

jjhelmus:
LGTM. Thanks for putting this together @bjlittle, merging.

PR #79, opened 2016-10-07
Update debug

jakirkham:
Merge master into debug.


jjhelmus:
I do not believe the debug branch of this repository is actually doing anything.  I do not recall why it is even around.  Would it make sense to remove it rather than update it?

jakirkham:
Fine with removal on my end.

PR #80, opened 2016-10-07
Update master

jakirkham:


jjhelmus:
What exactly is this doing?  It only seems to be adding commits and not changing files.  PR #77 already regenerated the feedstock.

jakirkham:
Was trying to get `debug` and `master` back in sync. If we delete `debug`, this can be closed.

PR #86, opened 2016-11-19
Add python-debug build feature

dfroger:
There are related disucssions [here](https://github.com/conda-forge/staged-recipes/pull/961/files) and [here](https://github.com/conda-forge/staged-recipes/issues/1593).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,9 +23,9 @@ before_install:
 
 install:
     - |
-      MINICONDA_URL="http://repo.continuum.io/miniconda"
+      MINICONDA_URL="https://repo.continuum.io/miniconda"
       MINICONDA_FILE="Miniconda3-latest-MacOSX-x86_64.sh"
-      curl -O "${MINICONDA_URL}/${MINICONDA_FILE}"
+      curl -L -O "${MINICONDA_URL}/${MINICONDA_FILE}"
       bash $MINICONDA_FILE -b
 
       source /Users/travis/miniconda3/bin/activate root
--- a/README.md
+++ b/README.md
@@ -33,7 +33,6 @@ conda search python --channel conda-forge
 ```
 
 
-
 About conda-forge
 =================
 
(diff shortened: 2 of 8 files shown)


edisongustavo:
This PR seems to change the default python to be compiled with `pydebug`.

I believe it is desired that both versions be compiled: with and without the pydebug feature.

dfroger:
Maybe I made the pull-request wrong in github, the intent was to add the code in a `python-debug` branch of the `python-feedstock` repo. (To avoid creating a new `python-debug-feedstock` repo, and simplify maintenance).

dfroger:
Doesn't seem to be possible to 
[pull request to a new upstream branch with GitHbub](http://stackoverflow.com/questions/9796370/github-pull-request-to-a-new-upstream-branch)

jjhelmus:
@dfroger Thanks for putting this together, I'm excited about getting a debug build on Python on conda-forge.  I went ahead and created a [python-debug](https://github.com/conda-forge/python-feedstock/tree/python-debug) branch in this feedstock which you should be able to direct a pull request against.  I think you may need to [rerender](https://github.com/conda-forge/conda-smithy#re-rendering-an-existing-feedstock) the feedstock to get the CI services to start working.

jakirkham:
That SO answer is outdated @dfroger . They changed this maybe a couple of months back. Just click edit near the title and it will provide the option to change the base branch.

dfroger:
@jjhelmus @jakirkham thanks, I changed the base branch to `python-debug`.

When I build the package on my machine, the package is successfully created, but then the test failed due to the feature. I tried to  [add the feature metapackage to the test: requires: section](https://github.com/conda/conda-build/issues/1026), but without success.

I'm going to rerender the feedstock.

jakirkham:
So re-render is definitely needed. However, I don't think that alone will solve the issue. I think we need a new metapackage that helps us enable/disable the debug feature for Python. Though if someone has a better idea feel free to share.

cc @msarahan @pelson

Issue #70, opened 2016-09-10
Python 3.5.2 build 3 does not place compiled Python files in __pycache__ directory


jjhelmus:
Ths conda-forge python 3.5.2_3 packages do not have the `.pyc` files in the standard library in `__pycache__` directory, rather they placed next to the `.py` directory.  The 3.5.2_2 build has the in the correct location.  

For example:

```
$ tar tvf ~/Downloads/osx-64-python-3.5.2-2.tar.bz2 | grep zipfile.*pyc
-rw-r--r--  0 travis staff    5000 Jul 25 20:38 lib/python3.5/test/__pycache__/test_zipfile64.cpython-35.opt-1.pyc
-rw-r--r--  0 travis staff    5000 Jul 25 20:38 lib/python3.5/test/__pycache__/test_zipfile64.cpython-35.pyc
-rw-r--r--  0 travis staff    5000 Jul 25 20:38 lib/python3.5/test/__pycache__/test_zipfile64.cpython-35.opt-2.pyc
-rw-r--r--  0 travis staff   44344 Jul 25 20:38 lib/python3.5/__pycache__/zipfile.cpython-35.opt-2.pyc
-rw-r--r--  0 travis staff   49866 Jul 25 20:38 lib/python3.5/__pycache__/zipfile.cpython-35.opt-1.pyc
-rw-r--r--  0 travis staff   49948 Jul 25 20:38 lib/python3.5/__pycache__/zipfile.cpython-35.pyc
-rw-r--r--  0 travis staff   75071 Jul 25 20:38 lib/python3.5/test/__pycache__/test_zipfile.cpython-35.opt-2.pyc
-rw-r--r--  0 travis staff   77015 Jul 25 20:38 lib/python3.5/test/__pycache__/test_zipfile.cpython-35.pyc
-rw-r--r--  0 travis staff   77015 Jul 25 20:38 lib/python3.5/test/__pycache__/test_zipfile.cpython-35.opt-1.pyc
```

vs.

```
$ tar tvf ~/Downloads/osx-64-python-3.5.2-3.tar.bz2 | grep zipfile.*pyc
-rw-r--r--  0 travis staff    4968 Sep  8 09:39 lib/python3.5/test/test_zipfile64.pyc
-rw-r--r--  0 travis staff    5044 Sep  8 09:38 lib/python3.5/test/__pycache__/test_zipfile64.cpython-35.opt-2.pyc
[...]

PR #87, opened 2016-11-21
Bump FD_SETSIZE on Windows

pitrou:


recipe/meta.yaml
```diff
@@ -11,6 +11,7 @@ source:
       - osx64-dist.patch                # [osx and x86_64]
       - win-find_exe.patch              # [win]
       - win-library_bin.patch           # [win]
+      - fd_setsize.patch                # [win]
 
 build:
   number: 3
```
ocefpaf: You need to bump the build number to get that change in.

recipe/fd_setsize.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/Modules/selectmodule.c b/Modules/selectmodule.c
+--- a/Modules/selectmodule.c
++++ b/Modules/selectmodule.c
+@@ -29,7 +29,7 @@
+    FD_SETSIZE higher before this; e.g., via compiler /D switch.
+ */
+ #if defined(MS_WINDOWS) && !defined(FD_SETSIZE)
+-#define FD_SETSIZE 512
++#define FD_SETSIZE 2048
```
ocefpaf: Can you write a quick sentence to why this change is needed?

pitrou:
@ocefpaf, thanks for the review comments. I think I've addressed them.

recipe/fd_setsize.patch
mrocklin: Why decide on number in particular?  Why not twice or half as many?

pitrou: Like any arbitrary limit, it just feels _a priori_ sufficient :-) Note there's a slight runtime cost as the fd array is runtime allocated; I think the cost is practically negligible but still.
By default:
```
>>> a, b= socket.socketpair()
>>> %timeit select.select([a], [a], [a])
The slowest run took 12.48 times longer than the fastest. This could mean that a
n intermediate result is being cached.
100000 loops, best of 3: 2.89 µs per loop
```
With this PR:
```
>>> %timeit select.select([a], [a], [a])
The slowest run took 24.39 times longer than the fastest. This could mean that a
n intermediate result is being cached.
100000 loops, best of 3: 3.52 µs per loop
```

pitrou:
I've rerendered in order to fix the CI bullds.

jjhelmus:
I'm not apposed to this change but I do think some discussion is warranted.  

The patches that are currently applied are either cosmetic or are needed to allow python to operate properly in a conda environment.  This patch is different, it changes the behavior of python in a manner that is not necessary for python to work in a conda environment (at least as far as I can tell).  

Additionally, this could create cases where a users script would work properly with the `conda-forge` python but not with the python from the `defaults` channel or the Python installed from the python.org installed.

I think making "improvement" to packages can be a slippery slope, who gets to decide what defaults should be changed or what trade-offs should be made.

Finally, has a similar change be suggested upstream, and if so how was it received?

pitrou:
@jjhelmus, the upstream issue reference is in the patch.

mrocklin:
Also issue raised for `default`  channel: https://github.com/ContinuumIO/anaconda-issues/issues/1241

jjhelmus:
Thanks for pointing out the upstream issue @pitrou, missed that link when I first looked at this.  I'm in favor of this change as this should keep this compatible with future versions of Python in the `defaults` channel.  

I think merging this change should wait until conda-forge moves to conda build 2.x to avoid creating a package missing the `__pycache__` directory, see #70.

jjhelmus:
@pitrou Now that the `__pycache__` issue is solved (see #70) do you want to rebase or create a new PR with a bump to FD_SETSIZE?  Python 3.6 already includes this change in the `master` branch but Python 3.5 still has the smaller size in the `3.5` branch.

jjhelmus:
This patch has been applied to Python 3.5 #109 and Python 3.6 #101.  Should this also be applied to Python 2.7, or can this issue be closed?

jjhelmus:
Closing issue.  Feel free to reopen if there is interest in applying this change to Python 2.7.

PR #96, opened 2016-12-07
Update prerelease channel to version 3.6.0rc1

jjhelmus:


jjhelmus:
[What's New In Python 3.6](https://docs.python.org/3.6/whatsnew/3.6.html).

PR #99, opened 2016-12-24
[Resend] Updating master to Python 3.6 release version.

Nelthorim:
Resend of #98.

PR #97, opened 2016-12-16
enable-loadable-sqlite-extensions tag added

xmnlab:
it is my first pull request in conda-forge.
any advice is welcome.

The sqlite3 module is not built with loadable extension support by default, because some platforms (notably Mac OS X) have SQLite libraries which are compiled without this feature. To get loadable extension support, you must pass –enable-loadable-sqlite-extensions to configure (https://docs.python.org/3/library/sqlite3.html#f1).

maybe before set this parameter would be necessary do some kind of validation.

PR #100, opened 2016-12-24
Updated to Python 3.6 release version.

Nelthorim:
Squashed commits from #99.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+* text=auto
+
+meta.yaml text eol=lf
+build.sh text eol=lf
+bld.bat text eol=crlf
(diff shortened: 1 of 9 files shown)


jjhelmus:
Thanks for jumping on this @Nelthorim!

Would really like to use `conda-build` 2.x for this build.  Also, incorporating #97 and #87 seems reasonable.  

I can work on a PR shortly if you would prefer.  I have a few other minor changes that would be good to add here also.

Nelthorim:
I'd prefer if you would do those changes, @jjhelmus. I have no personal experience with the conda build system, so I just made a the minimal change to update the prerelease branch into the release version of Python 3.6. It'd probably be faster and smoother if you did it than me figuring out how it works at this time.

Probably a bit ambitious for my first PR, but I really wanted to use Py3.6 in Windows without having stray Python installations around or PATH conflicts 😄

Nelthorim:
Replaced by #101

PR #101, opened 2016-12-24
Python 3.6.0

jjhelmus:


msarahan:
Weird default on Travis. Restarted.

jjhelmus:
Incorporated changes in #97 and #87 into this update.  @xmnlab and @pitrou can you review these changes and see if they meet the suggestions you made?

jjhelmus:
Is there any interest in updating the pinning to more recent versions with this new version of Python?  Specifically the following could be updated:

* readline `6.2*` -> `6.3*`
* sqlite `3.13.*` -> `3.15.*`
* tk `8.5.*` -> `8.6.*`

The version in the `defaults` channel does not seem to change the pinning.

jjhelmus:
Reverting the changes in #97 as these changes seem to cause [issues with the Linux build](https://circleci.com/gh/jjhelmus/python-feedstock/3?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

jjhelmus:
> Weird segfault on Travis.

This seems to be real, I can replicate this locally as well.  Tried a few changes and cannot figure out what is the root cause.  I see the same issue with the recipe in #100.

jakirkham:
If we update the pinnings, then they deviate from the pinnings that we are using with everything else.

msarahan:
@jjhelmus the segfault had something to do with conda build 2?

If so, it is probably the prefix replacement or rpath stuff messing things up.

scopatz:
This looks good to me FWIW

recipe/meta.yaml
```diff
@@ -1,24 +1,32 @@
...
       - omit-local-site-packages.patch  # [osx]
       - osx64-dist.patch                # [osx and x86_64]
       - win-find_exe.patch              # [win]
       - win-library_bin.patch           # [win]
+      # Adjust FD_SETSIZE to allow select() to wait on more connections
+      # simultaneously.  Desirable for serious use of Tornado on Windows.
+      # See also:
+      # - https://bugs.python.org/issue28708
+      # - https://github.com/zeromq/libzmq/issues/1165
+      - fd_setsize.patch                # [win]
```
jakirkham: Maybe we should be taking this on separately. Namely as a PR ( https://github.com/conda-forge/python-feedstock/pull/87 ). Doesn't feel like the sort of thing we should bury in a release.

pitrou:
Why isn't the feedstock versioned? The way this is done, it seems to suggest that conda-forge won't provide bugfixes for the Python 3.5 package anymore.

jakirkham:
There's already a `3.5` branch. `master` always contains the latest version.

That said, we may decide to stop supporting some versions/branches like `3.4`, but I expect we will be supporting `3.5` for a while.

pitrou:
I see. Thanks for correcting me.

ocefpaf:
@jjhelmus what is the status of this? Can we merge it?

jjhelmus:
> @jjhelmus what is the status of this? Can we merge it?

I think there is an issue with the OS X build when compiling with a long prefix environment.  Let me try to debug the issue in the next day or two.

jjhelmus:
No OS X issue!  I think this may be good to go if others concur.  

I can pull the fd_size patch out an submit is as a separate PR if that is preferred.

recipe/build.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 
-python ${RECIPE_DIR}/brand_python.py
+${SYS_PYTHON} ${RECIPE_DIR}/brand_python.py
+
+# Remove test data and ensurepip stubs to save space
+rm -rf Lib/test Lib/*/test
+rm -rf Lib/ensurepip
```
jjhelmus: Removing these files saves a significant amount of space, perhaps the same should be done in the Window build.

scopatz: +1, but maybe that can be done in a separate  pr

jjhelmus: I made the needed changes. @Nelthorim (thanks!) had them in his Python 3.6 PR so they were not hard to find.

jjhelmus:
I think this should be good to go now.  Can someone merge this or give a :+1: and I'll merge?

recipe/meta.yaml
jakirkham: Any thoughts on this point, @jjhelmus?

jjhelmus: I think it can be left it as the change is fine, but would be happy to pull the patch out and submit as a separate PR.

scopatz:
Woohoo!

PR #102, opened 2016-12-30
WIP: Cleanup

jakirkham:
Tries to cleanup how Python versions are handled in the recipe and build scripts to avoid having to update it in multiple places.

### DIFF
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -2,6 +2,16 @@ REM brand Python with conda-forge startup message
 python %RECIPE_DIR%\brand_python.py
 if errorlevel 1 exit 1
 
+python -c "import os; print(''.join(os.environ['PKG_VERSION'].split('.')[:2]))" > temp.txt
+if errorlevel 1 exit 1
+set /p MAJ_MIN_VER=<temp.txt
+if errorlevel 1 exit 1
+del /f temp.txt
+python -c "import os; print(os.environ['PKG_VERSION'].split('.')[0])" > temp.txt
+if errorlevel 1 exit 1
+set /p MAJ_VER=<temp.txt
+if errorlevel 1 exit 1
+del /f temp.txt
 
 REM Compile python, extensions and external libraries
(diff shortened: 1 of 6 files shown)


recipe/build.sh
```diff
@@ -25,5 +25,11 @@ fi
 
 make
 make install
-ln -s $PREFIX/bin/python3.5 $PREFIX/bin/python
-ln -s $PREFIX/bin/pydoc3.5 $PREFIX/bin/pydoc
+
+python -c "import os; print('.'.join(os.environ['PKG_VERSION'].split('.')[:2]))" > temp.txt
```
jjhelmus: I do not think you can use a `python` in this build script since python is not in the environment.  $SYS_PYTHON might be safe to use here.

recipe/bld.bat
```diff
@@ -2,6 +2,16 @@ REM brand Python with conda-forge startup message
 python %RECIPE_DIR%\brand_python.py
 if errorlevel 1 exit 1
 
+python -c "import os; print(''.join(os.environ['PKG_VERSION'].split('.')[:2]))" > temp.txt
```
jjhelmus: Can you use python here even if it is not installed in the build environment?

jjhelmus:
I think only idle3 is available on Linux not idle.  Looks to be causing the failing build.  

Also, now that Python 3.6.0 is out this should be re-targeted against the 3.5 branch.

recipe/bld.bat
jakirkham: Aren't we using it above for the branding?

jakirkham:
Odd that only on Linux there is no `idle` especially when it is on Mac. 😕

Should we add a symlink? It seems really valuable to have a non-versioned executable we can call too.

recipe/build.sh
jakirkham: We're using `python` above for the branding. What's the difference if we use it here?

mariusvniekerk:
Closing due to this never feasibly landing

PR #114, opened 2017-02-07
Keep test support (Python 3.6)

jakirkham:
Make sure `test.support` is kept on Python 3.6 as this is used by some packages to aid in running their test suites.

### DIFF
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -131,9 +131,12 @@ REM Populate the Lib directory
 del %PREFIX%\libs\libpython*.a
 xcopy /s /y %SRC_DIR%\Lib %PREFIX%\Lib\
 if errorlevel 1 exit 1
-REM Remove test data and ensurepip stubs to save space
+
+:: Remove test data to save space.
 rd /s /q %PREFIX%\Lib\test
 if errorlevel 1 exit 1
+
+:: Remove ensurepip stubs.
 rd /s /q %PREFIX%\Lib\ensurepip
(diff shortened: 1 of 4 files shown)


jjhelmus:
@jakirkham The test directory can be retained in the python package if downstream packages depend on it.  This may be easier than trying to pick and choose what files from that directory are needed.  The downside is a larger size but I'd take that over broken packages any day.

jakirkham:
I suppose that is an option. Though I think `test.support` (`test.test_support` in Python 2) is the only case where this might be used externally. This is because this module is part of the public API (please see refs below).

Note: This happened to me with a `backports` package, which is my motivation for these changes.

ref: https://docs.python.org/2.7/library/test.html#module-test.test_support
ref: https://docs.python.org/3.5/library/test.html#module-test.support
ref: https://docs.python.org/3.6/library/test.html#module-test.support

jakirkham:
Well this and the two other PRs for the other versions of Python are good to go on my end. Included a test to ensure `test` and `test.support` are present on all platforms. Please let me know if there is anything else.

jjhelmus:
LGTM, just waiting for the CIs to all go green then I'll merge these.  Thanks @jakirkham

jakirkham:
Looks like they just passed. 💚

jjhelmus:
Merging, thanks for adding this @jakirkham.

jakirkham:
Of course. Thanks for reviewing.

PR #128, opened 2017-03-19
Release 2.7.13

jakirkham:
Also drop `omit-local-site-packages.patch` as upstream has made this change to 2.7.13. So we no longer need to patch it ourselves. 🎉

xref: https://bugs.python.org/issue28440


jjhelmus:
LGMT

jakirkham:
Thanks @jjhelmus.

Issue #5, opened 2016-05-20
Python 2.6


jjhelmus:
Would be nice to build Python 2.6, although this may be a low priority.

[Source code for Python 2.6](https://www.python.org/download/releases/2.6.9/)

This might require some none standard setting in the CI script as conda-forge is not currently building Python 2.6 packages.

pelson:
Why would you do this to yourself? 😄  
I honestly don't think we should bother - 2.6 is long dead and should stay that way IMO... 😉

jakirkham:
If you want to maintain this @jjhelmus, I wouldn't stop you, but I don't know that I want to help support 2.6.

jakirkham:
Maybe I'm changing my mind. PR ( https://github.com/conda-forge/staged-recipes/pull/1153 ) and PR ( https://github.com/conda-forge/staged-recipes/pull/1157 ) add some backports for this. Though I would really consider these as one-offs not maintained feedstocks.

jakirkham:
Put together PR ( https://github.com/conda-forge/python-feedstock/pull/129 ) as an attempt to do this.

jakirkham:
This was done for Unix in PR ( https://github.com/conda-forge/python-feedstock/pull/129 ). Windows has not been done.

jakirkham:
Should we close? This got added for Unix. People are free to add Windows support if it is interesting.

PR #129, opened 2017-03-19
Python 2.6.9

jakirkham:
Fixes https://github.com/conda-forge/python-feedstock/issues/5

Attempts to build Python 2.6.9 mainly for historic value and people wishing to use it in CIs.

Note: This pulls in the history from 2.7.x as it shares more in common with the 2.6.9 build.


jakirkham:
The Windows failure appears related to a function called `find_best_ssl_dir` in `PCbuild/build_ssl.py`. FWR it is being given `..\\..` as the place to look for a directory that has filenames that start with `openssl-`. Not entirely sure why this was done in this manner. However, it will either need to be patched or overridden with some kind of option.

jakirkham:
I know this may not be the most interesting thing going on at conda-forge right now, but we do you think about this @jjhelmus? Right now Windows is bust, but I could skip it for now and we could have Mac and Linux packages.

jjhelmus:
Sorry been a bit busy the last few weeks @jakirkham.  I see some use for a Python 2.6 `conda-forge` release (`defaults` has one as well) but don't have the time to investigate the Windows issue at the moment.  I would be fine merging this with the Windows build skipped to get out Linux and MacOS packages.

jakirkham:
Thanks @jjhelmus. Have disabled Windows.

Do you know what we ended up deciding with `bzip2`, @jjhelmus? I remember discussing that we were using the static library before, but now we have issues as a dynamic library is also included in the `bzip2` package. This seems to be causing the failure on Linux. Not sure how macOS avoids it as it appears to be [linked]( https://travis-ci.org/conda-forge/python-feedstock/builds/224361142#L4969 ) to the same `libbz2` dynamic library.

jjhelmus:
There are likely two options for the bzip2 issue:

1. Link to the dynamic library and include bzip2 in the run requirements.
2. Delete the dynamic bzip2 library prior to the build so that the `bz2` module is linked statically.

There may be some issues with 1 if the bzip2 package is every updated in an environment as I believe conda uses the `bz2` module to unpack the package prior to linking.  Removing a dynamic library being used by a running process is a great way to cause a bus error.  So 2 is probably the safer option.  There is likely some compiler/linked switch that will force the bzip2 library to be statically linked but I think removing the dynamic library would require less work.

jakirkham:
Perfectly fine with deleting the bzip2 dylib/so as part of the build script. Have pushed a fix along those lines to this branch. If it works, we can port it to the other Python branches as well.

jakirkham:
Took some tweaking, but this seems to be doing the right thing now. What do you think?

jjhelmus:
LGTM, thanks for putting this together @jakirkham

PR #131, opened 2017-04-02
Stackless Python 2.7.13 [create new branch "2.7-stackless]

akruis:


akruis:
Currently all builds fail. Linux and OS-X build, but tests fail, because the meta package "stackless" is not yet available.

But Windows fails to untar the source. Stackless comes as a tar.xz archive. I already added package "xz" as a build requirement, because it provides the command "unxz" (at least on linux). I'm going to reproduce this on my Windows box tomorrow.

This is the relevant part of the AppVeyor log:
```
The following NEW packages will be INSTALLED:
    vs2015_runtime: 14.0.25420-0 conda-forge
    xz:             5.2.2-0      conda-forge
Source cache directory is: C:\bld\src_cache
Downloading source to cache: stackless-2713-export.tar.xz
Downloading http://www.stackless.com/binaries/stackless-2713-export.tar.xz
Success
unxz is required to unarchive .xz source files.
```

jakirkham:
Having Stackless Python in conda-forge does sound really cool. I'd like to reach out to folks at Continuum to see if they have any thoughts on how we approach this.

cc @msarahan @mingwandroid @ccordoba12 @mcg1969

akruis:
@jakirkham I'm glad you like the idea of getting Stackless into conda-forge. 

Meanwhile I created a tar.bz2 source archive on stackless.com and modified the recipe to use it. Now all three build bots build the package and fail when they try to install the stackless meta package (https://github.com/conda-forge/staged-recipes/pull/2714).

jakirkham:
So, as I don't know as much about the inner workings of Stackless, would we need to rebuild C extension packages to work with Stackless?

jjhelmus:
I'm wondering if a different package name shouldn't be used for stackless Python rather than a feature?  I do not believe conda provides a mechanism for selecting features without installing a meta-package.  Users might be suprised if `conda install -c conda-forge python` did not install CPython.  For better-or-worse most Python users equate Python == CPython.

akruis:
@jjhelmus Stackless-Python is a C-Python with some additions. This makes it hard, to install C-Python and Stackless in the same conda environment. There are file name collisions: `python27.dll` for instance. The Stackless installation would also need its own `Lib` and `include` directory. All this can be resolved, but it gets nasty.

Now, if we create a Stackless package, that installs Stackless besides C-Python, how to select the right binary for an application? If the application uses Stackless features, you must use the “python”-binary provided by the Stackless package. But the user of an application might not know, that the application contains a module, that requires Stackless Python.

Therefore I would like to use the conda feature mechanism. The recommended way to track a feature is to use a meta package. Unless the meta package gets installed, you never get Stackless Python and most users of conda-forge won't ever notice, that Stackless Python is available. Of course the meta package must not be installed by default. But I don't think, that it is common to install all conda-forge packages.

mattip:
I think this can be closed.

PR #140, opened 2017-04-30
Fix bz2 linking and rerender 2.3.0 - Python 2.7

jakirkham:
* Use static linking with bzip2.
* Re-render with conda-smithy 2.3.0.

PR #141, opened 2017-04-30
Apply bzip2 shared library fix and rerender recipe - Python 2.7

jjhelmus:


jakirkham:
Oops, had just added PR ( https://github.com/conda-forge/python-feedstock/pull/140 ). Feel free to close it though if you would rather. Thanks for doing these.

jjhelmus:
@jakirkham Just saw that as well.  Thanks for putting together the bzip2 fix, I borrow it for these PRs.

jjhelmus:
I do not believe the failing Windows build is the result of the rerender, rather `%PYTHON%` is not defined when building python, conda/conda-build#1960.  Let me think about the best method for solving this.  It might be necessary to provide a full path to the newly built Python interpreter in the `bld.bat` script.

jjhelmus:
I think this is good to go.  Anyone care to review the changes?

recipe/bld.bat
```diff
@@ -150,9 +151,9 @@ if errorlevel 1 exit 1
 
 REM bytecode compile the standard library
 
-%PYTHON% -Wi %STDLIB_DIR%\compileall.py -f -q -x "bad_coding|badsyntax|py3_" %STDLIB_DIR%
+%PREFIX%\python.exe -Wi %PREFIX%\Lib\compileall.py -f -q -x "bad_coding|badsyntax|py3_" %PREFIX%\Lib
```
jakirkham: After looking everything, this change seems fine as these paths all end up being the same in the end.

PR #139, opened 2017-04-30
Apply bzip2 shared library fix and rerender recipe - Python 3.5

jjhelmus:

PR #142, opened 2017-05-10
Fixes for Windows GCC interop needed by RPy2 and (potentially) CVXOPT

mingwandroid:


mingwandroid:
Here you go @ocepaf

jjhelmus:
LGTM.  Wanted to check if these new patches should be applied on all platforms or just Windows?

mingwandroid:
All patches on all platforms always.

jjhelmus:
Thanks for the updates @mingwandroid

recipe/meta.yaml
```diff
@@ -19,9 +19,11 @@ source:
     # - https://bugs.python.org/issue28708
     # - https://github.com/zeromq/libzmq/issues/1165
     - fd_setsize.patch                # [win]
+    - mingw-w64--DMS_WINXX.patch
+    - win-msc_ver-1900.patch
 
 build:
-  number: 1
+  number: 2
```
isuruf: Github is playing with us. This was `2` before and there should be no change here. Build number `2` was set in this PR https://github.com/conda-forge/python-feedstock/pull/139/files.

cc @ocefpaf

ocefpaf: Thanks @isuruf! That explains the `rpy2` problem. I'll send a PR to bump it now.

PR #143, opened 2017-05-10
Fixes for Windows GCC interop needed by RPy2 and (potentially) CVXOPT

mingwandroid:
Here you go @ocefpaf

### DIFF
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -21,7 +21,7 @@ cd ..
 
 
 REM Populate the root package directory
-for %%x in (python36.dll python.exe pythonw.exe) do (
+for %%x in (python36.dll python3.dll python.exe pythonw.exe) do (
     copy /Y %SRC_DIR%\PCbuild\%BUILD_PATH%\%%x %PREFIX%
     if errorlevel 1 exit 1
 )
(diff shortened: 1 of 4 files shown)


mingwandroid:
OK, done:

https://github.com/conda-forge/python-feedstock/pull/142 and https://github.com/conda-forge/python-feedstock/pull/143

ocefpaf:
Thanks @mingwandroid!

jjhelmus:
Thanks for this @mingwandroid. Is there a reason for replicating the patch files in the root directory and the recipe directory?  Also should these patches be applied on all platforms or just Windows?

mingwandroid:
I have a problem with the idea of patching on only some platforms. That means that your patch is possibly not upstreamable because it may break other platforms. Patches should never break other platforms and if they do they need to be reworked until they don't.

If you were to attempt to cross compile Python modules on Linux using mingw-w64 GCC then this patch would be necessary there too.

mingwandroid:
They shouldn't be in the root, I'll force push in a mo.

jjhelmus:
LGTM, will merge once the CI are green.  Thanks @mingwandroid

mingwandroid:
Oh I have another fix incoming.

PR #145, opened 2017-05-14
Fix ABI breakage (issue29943)

mingwandroid:


jjhelmus:
Thanks @mingwandroid for the fix.  

In case anyone is interested in this issue it is discussed in https://bugs.python.org/issue29943.  Briefly Python 3.6.1 broke ABI compatibility with Python 3.6.0 in a manner in which many C extensions build on Python 3.6.1 will fail to import in Python 3.6.0.  Extensions build on Python 3.6.0 will work on 3.6.0 and 3.6.1.  This change will fix the ABI compatibility issue and result in 3.6.1 extensions working on both 3.6.1 and 3.6.0.

jschueller:
Note : this is not fixed for windows as the python dll is not recompiled.

jjhelmus:
> Note : this is not fixed for windows as the python dll is not recompiled.

Can you provide some more details on this?  We are building the python dll in this recipe and it should include the fix in this PR.

PR #146, opened 2017-05-22
build Python 2 with UCS-4 on darwin

minrk:
matches linux

Are there any downstream things this would break? It makes Python 2 unicode much more sensible (as it is on Linux already).

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -26,7 +26,7 @@ if [ `uname` == Darwin ]; then
     ./configure \
         --enable-ipv6 \
         --enable-shared \
-        --enable-unicode=ucs2 \
+        --enable-unicode=ucs4 \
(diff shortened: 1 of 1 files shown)


minrk:
Here's an issue suggesting this is probably not a good idea: https://github.com/pyenv/pyenv/issues/257

jjhelmus:
@minrk In addition to the issue you linked, the macOS Python 2.7 in the `defaults` channel is UCS-2.  Breaking compatibility with this version is undesirable as it could result in incompatibilities in many of the compiled C extensions already available.

minrk:
Indeed, I forgot that the underlying unicode implementation is an ABI/API issue at the C-level, and that's why Python 2 on darwin users are stuck with UCS-2 ~forever.

PR #148, opened 2017-07-19
update to 3.6.2

ocefpaf:
@jjhelmus can you take a look at this one?

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "3.6.1" %}
+{% set version = "3.6.2" %}
 
 package:
   name: python
@@ -7,7 +7,7 @@ package:
 source:
   fn: Python-{{ version }}.tgz
   url: https://www.python.org/ftp/python/{{ version }}/Python-{{ version }}.tgz
-  sha256: aa50b0143df7c89ce91be020fe41382613a817354b33acdc6641b44f8ced3828
+  sha256: 7919489310a5f17f7acbab64d731e46dca0702874840dadce8bd4b2b3b8e7a82
   patches:
     - omit-local-site-packages.patch  # [osx]
     - osx64-dist.patch                # [osx and x86_64]
@@ -24,7 +24,7 @@ source:
     - fix-ABI-issue29943.patch
 
 build:
-  number: 3
+  number: 0
   # Windows has issues updating python if conda is using files itself.
   # Copy rather than link.
   no_link:
@@ -62,6 +62,8 @@ test:
     - 2to3 -h
     - pydoc -h
     - python3-config --help  # [unix]
+    - conda inspect linkages -p $PREFIX python  # [not win]
+    - conda inspect objects -p $PREFIX python  # [osx]
 
 about:
   home: http://www.python.org/
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -22,6 +22,7 @@ source:
     - mingw-w64--DMS_WINXX.patch
     - win-msc_ver-1900.patch
     - fix-ABI-issue29943.patch
+    - do-not-download-externals.patch  # [win]
 
 build:
   number: 0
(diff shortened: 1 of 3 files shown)


ocefpaf:
Windows builds are timing out :unamused: 

They seem to hang at:

```
(C:\bld\python_1500472545057\_b_env) C:\bld\python_1500472545057\work\Python-3.6.2>cd PCbuild 
(C:\bld\python_1500472545057\_b_env) C:\bld\python_1500472545057\work\Python-3.6.2\PCbuild>call build.bat -e -p Win32 
Checking for installed python...
Downloading nuget...
Invoke-WebRequest : The operation has timed out.
At line:1 char:1
+ Invoke-WebRequest https://aka.ms/nugetclidl -OutFile 'C:\bld\python_1 ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:Htt 
   pWebRequest) [Invoke-WebRequest], WebException
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShe 
   ll.Commands.InvokeWebRequestCommand
 
Installing Python via nuget...
```

jjhelmus:
Looks like the build systems for windows changed, something in `build.bat` is now trying to install python from nuget.  I don't see any mentioned of this in the release notes, but I could be missing something.  I will make some time today to investigate and find a solution.

jjhelmus:
Looks like the windows build change was from [bpo-30450](https://bugs.python.org/issue30450), 

> The build process on Windows no longer depends on Subversion, instead pulling external code from 
> GitHub via a Python script. If Python 3.6 is not found on the system (via py -3.6), NuGet is used to 
> download a copy of 32-bit Python.

jjhelmus:
Within the new build process the script, [get_externals.bat](https://github.com/python/cpython/blob/v3.6.2/PCbuild/get_externals.bat), uses NuGet to download Python 3.6 which in turn is used to run the [get_external.py](https://github.com/python/cpython/blob/v3.6.2/PCbuild/get_external.py) script.  This downloads and unpacks the seven source zip files from [python/cpython-source-deps](https://github.com/python/cpython-source-deps/releases) and a zip file containing the nasm binary from [python/cpython-bin-deps](https://github.com/python/cpython-bin-deps/releases).  

I think it would be possible to use `curl` and `7za` to perform this unpacking and then patch the build script to avoid calling `get_externals.bat`.  Do this seem reasonable to other or are there other suggestions?

Another option is to provide these dependencies through conda packages, #8. This requires modification of the Visual Studio files, see #19, and would be a more involved process.

ocefpaf:
> I think it would be possible to use `curl` and `7za` to perform this unpacking and then patch the build script to avoid calling `get_externals.bat`.  Do this seem reasonable to other or are there other suggestions?

This looks like the "easiest" solution at the moment.

> Another option is to provide these dependencies through conda packages, #8. This requires modification of the Visual Studio files, see #19, and would be a more involved process.

IMO this looks like a nice solution for the longer term.

PS: I cannot work on either of them b/c I don't have a Windows machine and iterating on AppVeyor is time consuming.

Feel free to merge/close/pick the commit, etc from this PR. Hopefully someone a Windows machine can help us fix this later. Thanks @jjhelmus for the explanations!

jjhelmus:
Green!  Thanks @ocefpaf.  Merging.

ocefpaf:
Thank you for fixing the hard part ;-p

recipe/meta.yaml
```diff
@@ -22,9 +22,10 @@ source:
     - mingw-w64--DMS_WINXX.patch
     - win-msc_ver-1900.patch
     - fix-ABI-issue29943.patch
+    - do-not-download-externals.patch  # [win]
```
jakirkham: Looking at the patch, would we be able to just set `IncludeExternals` to `false` and not need this patch or would that have other effects later on in that file?

jjhelmus: From what I recall, setting IncludeExtenals to false results in some modules in the standard library which depended on these external libraries not being built.  This may no longer be the case.

PR #151, opened 2017-10-04
2.7.14

ocefpaf:
See https://www.python.org/downloads/release/python-2714/


jjhelmus:
Thanks @ocefpaf, looks great.

recipe/run_test.py
```diff
@@ -83,7 +80,6 @@
 import strop
 import time
 import test
-import test.test_support
```
jakirkham: Why was this test dropped?

jakirkham: Turns out some content got shuffled around in Python 2.7.14. Fixing in PR ( https://github.com/conda-forge/python-feedstock/pull/168 ).

PR #160, opened 2017-12-07
Fix find_library so that it looks in sys.prefix/lib first

tacaswell:
Conflicts:
	recipe/meta.yaml
	  Many more patches are applied to AnacondaRecipes version


jakirkham:
Not sure if you did, but we might need [this patch]( https://github.com/AnacondaRecipes/python-feedstock/blob/master-3.5/recipe/0017-Fix-find_library-so-that-it-looks-in-sys.prefix-lib-.patch ) instead of the one in `master`.

tacaswell:
Ah, yup that makes sense (and I'll stop trying to hand edit diff files....)

jakirkham:
Sorry. Should have pointed that out sooner. 😳

tacaswell:
No worries!

PR #174, opened 2018-03-30
Use our build of nasm

jakirkham:
Instead of downloading the binary for `nasm` from CPython. Just use our build of it.

### DIFF
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -15,11 +15,7 @@ for %%x in (xz-5.2.2 tk-8.6.6.0 tix-8.4.3.6 tcl-core-8.6.6.0 sqlite-3.21.0.0 ope
     if errorlevel 1 exit 1
 )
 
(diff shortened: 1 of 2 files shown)


jjhelmus:
LGTM. Should the build number be bumped to trigger the creation of a new package or should we wait until another change since there should be no difference in the produced binary?

jakirkham:
Either way. Do you have a preference?

jjhelmus:
Merging as is. I want to include a patch for better mingw support today which will require a rebuild.

jjhelmus:
thanks for the update @jakirkham

jakirkham:
Sounds good. Thanks @jjhelmus.

Would it be worthwhile to do something similar for the other Python versions? Interested in avoiding downloading binaries that we haven't built if possible. Though it doesn't look like it would be as easy in the other cases.

PR #176, opened 2018-03-30
add patch for windows gcc interop

jjhelmus:
passes -DMS_WIN32 or -DMS_WIN64 to gcc on Windows

patch from:
https://github.com/AnacondaRecipes/python-feedstock/tree/master-2.7


jjhelmus:

jakirkham:
Thanks @jjhelmus. Remember seeing this come up in the context of `backports.lzma` as well. Seem to recall we don't need this for Python 3. Is that correct?

jakirkham:
Ah ok. Missed that. Thanks.

jakirkham:
Does CPython have any thoughts on incorporating this change?

PR #183, opened 2018-06-28
Go to 3.7!

jschueller:
Not sure if all patches are still relevant though, the one about abi issue 29943 seem useless now, but I dont kwow about others.


mariusvniekerk:
We may want to bring this closer to the recipe here https://github.com/AnacondaRecipes/python-feedstock/blob/master-3.6/recipe/meta.yaml

mariusvniekerk:

jakirkham:
There's a bunch of manual pinnings in here that can be dropped (most of which are outdated). Would be good to use the new `compiler` syntax as well. Needs a re-render after doing that to add pinnings from `conda-forge-pinning`.

ref: https://conda-forge.org/docs/meta.html#pinning-packages
ref: https://conda-forge.org/docs/meta.html#compilers

jschueller:

jschueller:

ocefpaf:
The number of patches in the Anaconda version is 16!

https://github.com/AnacondaRecipes/python-feedstock/tree/master/recipe

We already have 11 patches which is 11 more than what I would like :smile: 

Maybe @mingwandroid or @jjhelmus can help us understand if all of those are needed of if there are some compiler specific stuff there.

mingwandroid:
We need to try to remove some, 0016-, the most recent one for example. Most of them have pretty good comments in them. We can try to drip feed them from our fork a PR at a time if you want to split the discussion and review process up that way? Each one is fairly independent.

jschueller:

ocefpaf:
> We need to try to remove some, 0016-, the most recent one for example. Most of them have pretty good comments in them. We can try to drip feed them from our fork a PR at a time if you want to split the discussion and review process up that way? Each one is fairly independent.

Don't get me wrong. I'm OK with patching in general. Just worried about the maintainability. I'll read the comments to get better informed on them and come back with questions (if I have any). Thanks!

jschueller:
the windows build fails, the externals dependencies (openssl, tk, xz, ...) do not seem to be built although I updated the required versions, any ideas ?

ocefpaf:
> the windows build fails: the externals dependencies (openssl, tk, xz, ...) do not seem to be built although I updated the required versions, any ideas ?

Take a look at https://github.com/AnacondaRecipes/python-feedstock/tree/master/recipe
looks like they are getting those deps in a different way.

jschueller:

jschueller:

mariusvniekerk:
Could we merge the version from anaconda recipes? I'd love for us to get more in sync

ocefpaf:
> Could we merge the version from anaconda recipes? I'd love for us to get more in sync

Even though I agree with @mariusvniekerk here and we should try to sync as much as we can, sometimes the "git" burden to do that may drive away contributors and we should be careful when asking that. @jschueller if you are git-savvy and willing to fetch the commits there and merge than here it will be greatly appreciated, but unless someone else is willing to do so I think we should not block this PR waiting for it.

mingwandroid:
At AD, I personally don't hugely care about credit and keeping commits for that end.

Sure good rebasing is best when possible but I wouldn't get too worried about us rebasing on top of cf, we can manage that.

The idea is for our stuff to become reduced to the essential differences between the ecosystems and in many cases vanish.

mariusvniekerk:
@jschueller seems the windows build is failing due to some tk things

https://github.com/jschueller/python-feedstock/blob/99923e179bdf65e7f21b1e4890d459dc74b30cc1/recipe/bld.bat#L98-L105

This fails

PR #189, opened 2018-07-25
update recipe to the new format

ocefpaf:


ocefpaf:
@isuruf can you take a quick look at this one?

isuruf:
LGTM. Needs a rerender

ocefpaf:
Looks like the webservice did not re-render this properly in https://github.com/conda-forge/python-feedstock/pull/189#issuecomment-407894449

recipe/meta.yaml
```diff
@@ -29,20 +27,23 @@ build:
```
isuruf: Add  `skip: True   # [win and vc!=9]`

isuruf: Also
```
track_features:
- vc9               # [win]
```

can be removed I think. We removed it for other python versions

ocefpaf: Thanks! I had those when I copied the recipe from AnacondaRecipes but ended up reverting and doing a bad copy-n-pasta.

ocefpaf:
Windows is failing with:

```
Using msbuild  (found in the PATH)

(C:\bld\python_1532577008425\_build_env) C:\bld\python_1532577008425\work\PCbuild>msbuild
  "C:\bld\python_1532577008425\work\PCbuild\pcbuild.proj" /t:Build  /nologo
  /v:m /p:Configuration=Release /p:Platform=x64 /p:IncludeExternals=true
  /p:IncludeSSL=true /p:IncludeTkinter=true /p:IncludeBsddb=true           

C:\bld\python_1532577008425\work\PCbuild\pcbuild.proj(3,18): error MSB4066:
  The attribute "Label" in element <PropertyGroup> is unrecognized.
```

scopatz:
This might be an  issue in the actual Python source. Or you might just need to update .NET to v4.0: https://stackoverflow.com/questions/5107757/using-msbuild-with-vs2010-generated-vcxproj-file-as-target-error-msb4066-at

ocefpaf:
> Or you might just need to update .NET to v4.0

How can we do that in AppVeyor?

scopatz:
You could try a different worker image, maybe: https://www.appveyor.com/docs/build-environment/#build-worker-images

scopatz:
the 2015 aand 2017 images seem to have v4.0+ images, while 2013 does not.

isuruf:
python 2.7 needs VS2010 or newer msbuild.exe to build the project even though VS2008 compilers are used.

In python 2.7.14 or before what pcbuild.bat did was activate VS2008 environment and then activated VS2015 or VS2010 (whichever was found first). See https://github.com/python/cpython/blob/2.7/PCbuild/env.bat#L14

In python 2.7.15 this behaviour was changed to use the msbuild.exe on the PATH if it was activated. (https://github.com/python/cpython/commit/986b7ffc650919b3022ccaa458a843bb8a95d2bd#diff-ab03c5bd5b42c0648e61e3553dcbea19L111)

Since vs2008 compiler activates the VS2008 environment, msbuild.exe from VS2008 is at the front, which is older than needed.

Other question was how were we able to build python 2.7.15 before. Due to a foresight, VS2015 was used in the activation. This is okay since the pcbuild doesn't actually use the compiler from the activated environment.

TL;DR: `call PCbuild\env.bat` which would setup the newer environment.

PR #190, opened 2018-08-15
Add backport from anaconda-recipes

mariusvniekerk:
Checklist
* [ ] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy`


mariusvniekerk:
@jjhelmus Ready for review

jjhelmus:
After thinking about this more I think there are some additional modifications that will be needed to allow Python to work properly with the new compilers.  There are a number of adjustments in the [build.sh file in AnacondaRecipes](https://github.com/AnacondaRecipes/python-feedstock/blob/master-2.7/recipe/build.sh) that effect the contents of the `_sysconfig_*.py` that will be needed.  Additionally, a generic [_sysconfig.py file](https://github.com/AnacondaRecipes/python-feedstock/tree/master-2.7/recipe/sysconfigdata) is included in the package which is used when not using the conda provided compilers.

I will put together a PR with these changes this weekend.

PR #192, opened 2018-08-18
Unbunddle readline from main python package as it is covered by GPL

### LINKED ISSUE #191: Does readline make the package GPL?
I'm pretty sure compiling with readline makes the package GPL.

https://github.com/conda-forge/python-feedstock/blob/576821dc17bd47ad0819d4fd60ce69a27bc27f42/recipe/meta.yaml#L118

Is there an exception I'm missing?

Can we compile python without readline support? Does it really matter that it doesn't have readline support?

Also, it seems that the sqlite package, that python depends on also depends on readline: https://github.com/conda-forge/sqlite-feedstock/issues/29

hmaarrfk:
I need some help to finish this.
Fixes #191

Basically, python will dynamically link any module called `readline` at run time.
There seem to be 2 options:
`libedit` which is shipped with OSX, and `readline` which is shipped with linux.
Finally, issue #191 links to a python issue where they have been trying to make compiling more friendly with libedit, it might be worth taking a look into it if we care.

Without readline, you don't have functionality like the "up-arrow".
By including readline, we should really advertise the package as GPL and add the appropriate license.

For reference, python never releases unix binaries themselves, so we can't use them as a reference.

Ultimately, it is a matter of "advertising the correct license". In the package, it is stated the binary is under PSF. Unfortunately, the binary that is compiled with readline is protected under GPL. libpython is not. Therefore, we must either advertise the product as GPL, and give it i a GPL license, or we must break off the GPL dependency into its own thing.

This PR creates an output `python-readline` that is GPLv3. Unfortunately, I don't know how to make the base `python` package include everything else. This is where I would need help from @conda-forge/core 

Finally, I recommend that https://github.com/conda-forge/sqlite-feedstock/pull/31 be worked on an pulled in first. This package depends on `libsqlite` that I split off from `sqlite` for the same reason. That package was smaller, so manually listing the files in `libsqlite` was a *viable* option.

Note: everything hidden in these details pertains to a previous solution of mine where by the readline module was simply not compiled. I don't think that is necessary anymore. I think it is fine to compile in the same step, but just to ship them as two separate packages
<details>
This depends on https://github.com/conda-forge/sqlite-feedstock/pull/30 being pulled in. I don't know how to put that in the dependencies. 

If sqlite doesn't get accepted first, then python automatically detects readline and compiles with it.

The relevant lines of the `./configure` output are 
```
checking how to link readline libs... none
checking for rl_pre_input_hook in -lreadline... no
checking for rl_completion_display_matches_hook in -lreadline... no
checking for rl_resize_terminal in -lreadline... no
checking for rl_completion_matches in -lreadline... no
checking for append_history in -lreadline... no
```
</details>.

~~Not sure how ipython does it, but ipython completion is still working :S. I'll look into further.~~
ipython seems to have its own editing code independent from readline.


hmaarrfk:
I'll rebase if you don't want the rerender as part of this PR.

hmaarrfk:
In a clean environement:

```
➤ conda activate p2                                                                               (base) 
set: Warning: $PATH entry "/home/mark/miniconda3/envs/p2/bin" is not valid (No such file or directory)
                                                                              [ 0s156 | Aug 18 11:46AM ]
mark@xps:~
➤ conda install python --use-local                                                                  (p2) 
Solving environment: done

## Package Plan ##

  environment location: /home/mark/miniconda3/envs/p2

  added / updated specs: 
    - python


The following NEW packages will be INSTALLED:

    bzip2:           1.0.6-h470a237_2        conda-forge
    ca-certificates: 2018.8.13-ha4d7672_0    conda-forge
    certifi:         2018.4.16-py37_0        conda-forge
    libedit:         3.1.20170329-haf1bffa_1 conda-forge
    libffi:          3.2.1-hfc679d8_4        conda-forge
    libgcc-ng:       7.2.0-hdf63c60_3        conda-forge
    libstdcxx-ng:    7.2.0-hdf63c60_3        conda-forge
    ncurses:         6.1-hfc679d8_1          conda-forge
    openssl:         1.0.2o-h470a237_1       conda-forge
    pip:             18.0-py37_1             conda-forge
    python:          3.7.0-hb7eb50a_0        local      
    setuptools:      40.0.0-py37_1           conda-forge
    sqlite:          3.24.0-h640cc1b_1       local      
    tk:              8.6.8-0                 conda-forge
    wheel:           0.31.1-py37_1           conda-forge
    xz:              5.2.4-h470a237_1        conda-forge
    zlib:            1.2.11-h470a237_3       conda-forge

Proceed ([y]/n)? 
```
no longer pulls in readline.

After installing `ipython`:

```
➤ conda list                                                                                        (p2) 
# packages in environment at /home/mark/miniconda3/envs/p2:
#
# Name                    Version                   Build  Channel
backcall                  0.1.0                      py_0    conda-forge
bzip2                     1.0.6                h470a237_2    conda-forge
ca-certificates           2018.8.13            ha4d7672_0    conda-forge
certifi                   2018.4.16                py37_0    conda-forge
decorator                 4.3.0                      py_0    conda-forge
ipython                   6.5.0                    py37_0  
ipython_genutils          0.2.0                      py_1    conda-forge
jedi                      0.12.1                   py37_0  
libedit                   3.1.20170329         haf1bffa_1    conda-forge
libffi                    3.2.1                hfc679d8_4    conda-forge
libgcc-ng                 7.2.0                hdf63c60_3    conda-forge
libstdcxx-ng              7.2.0                hdf63c60_3    conda-forge
ncurses                   6.1                  hfc679d8_1    conda-forge
openssl                   1.0.2o               h470a237_1    conda-forge
parso                     0.3.1                      py_0    conda-forge
pexpect                   4.6.0                    py37_0  
pickleshare               0.7.4                    py37_0  
pip                       18.0                     py37_1    conda-forge
prompt_toolkit            1.0.15                   py37_0  
ptyprocess                0.6.0                    py37_0  
pygments                  2.2.0                      py_1    conda-forge
python                    3.7.0                hb7eb50a_0    local
setuptools                40.0.0                   py37_1    conda-forge
simplegeneric             0.8.1                      py_1    conda-forge
six                       1.11.0                   py37_1    conda-forge
sqlite                    3.24.0               h640cc1b_1    local
tk                        8.6.8                         0    conda-forge
traitlets                 4.3.2                    py37_0  
wcwidth                   0.1.7                      py_1    conda-forge
wheel                     0.31.1                   py37_1    conda-forge
xz                        5.2.4                h470a237_1    conda-forge
zlib                      1.2.11               h470a237_3    conda-forge
```
I checked the packages, they are all BSD, MIT, ISC, or Apache, except for libgcc which has an exception for linking

hmaarrfk:
if this PR gets accepted, we could make `python` depend on `libsqlite` which doesn't have the `readline` dependency
https://github.com/conda-forge/sqlite-feedstock/pull/31

hmaarrfk:
@msarahan this is the particular recipe I care about.

I would like to have the python distribution be clearly GPL free. as readline can be dynamically linked into any python program if it is included by default, it creates a case that any project that runs on python, and depends on this package may be GPL.

Maybe i'm doing something wrong.

hmaarrfk:
Maybe I'm missing the `activate in script` flag
https://github.com/AnacondaRecipes/boost-feedstock/blob/master/recipe/meta.yaml#L118

scopatz:
Hi @hmaarrfk - I think that the potential for this to break downstream packages is very large, as there is a lot of code out there that expects Python to have readline. I think this needs to be covered under the nogpl-namespace again.

jjhelmus:
Rather than splitting the python package perhaps creating a python build variant which was built without readline would be an option.  The [selectable_pkg_variants_example repository](https://github.com/jjhelmus/selectable_pkg_variants_example) provides some details on one way that this could be arranged.  The basic idea is that there we be parallel `python` package which did not include the readline linking which could be selected using `conda install python-gplfree`.  This would install the `python` package without readline.

scopatz:
@jjhelmus - yes, I think this exactly what is needed, if we want to have a parallel stream on non-gpl packages.

hmaarrfk:
@scopatz I think if you decide to go down that route, it is a ticking time bomb, in terms of GPL violation letters arriving at Anaconda's door (as they have similar to conda-forge's). You are right that this is a big change, and might require a mass rebuild to identify what packages actually depend on readline.

Those packages that do use readline functionality should be labeled as GPL. They SHOULD break, so they can be notified that they are depending on a GPL'ed package and finially abide by the terms of readline's license.

I don't really think many packages depend on readline for their core functionality. Normal Python for MacOX doesn't ship with readline. Windows on anaconda/conda-forge doesn't ship with readline. readline is mostly a user experience thing, and by installing `python-readline` users can regain that experience. Software that depends on `readline` functionality being backed it probably counts as a "combined work".

For a long time, python was compiled without readline support and people kinda just lived with it.

That said, this all goes back to your [interpretation](https://github.com/conda-forge/sqlite-feedstock/pull/31#discussion_r211633790) of what is actually included when you use an API.
In C/C++, you literally **include** a whole file with gcc. The GNU and FSF interpret that as copying in a file. Personally, I wouldn't want to mess with GPL violation letters.

hmaarrfk:
@jjhelmus build variants are interesting, but I don't they are the correct solution because:

1. They don't stop the potential virality of the GPL license
2. The build variant in this case seems overkill as it is only a single file we are interested in moving out of the distribution. A split package seems like a much more straightforward solution, that wouldn't necessitate creating a whole separate ecosystem built on that variant.


If you would agree to split the package. If split packages with conda are not possible, then I can simply create a new recipe that would compile this specific module and install it in the correct location.

This all depends A. your interpretation of GPL, and B. on your decision on what should be included in the package named `python`. I think it should be GPL-free.

scopatz:
> They SHOULD break, so they can be notified that they are depending on a GPL'ed package and finially abide by the terms of readline's license.

I disagree, because most Python programs that depend on readline are interpreted (not extension modules) and therefore subject to the GPL license.  Also, to be clear what we are talking about breaking here is the Jupyter console, IPython, and (of concern to me) xonsh. This is a large userbase that should not just break, when there are alternatives that involve fixing it.

> In C/C++, you literally include a whole file with gcc

Yes, header files are included, but these (esp for readline) are APIs, which are not copyrightable.

> The GNU and FSF interpret that as copying in a file. 

Yes they do, but the FSF doesn't get to decide what the law is.  

I also think there might be some confusion over what the point of the `license` and related field mean.  I think that you are interpreting this to mean what the license of the package is.  This can change with dependencies as noted by linking. I have interpreted the `license` field to mean what license the project distributes itself under.

scopatz:
> B. on your decision on what should be included in the package named python. I think it should be GPL-free.

Yeah, I think that it should follow from whatever the build instructions in the official documentation are. This ends up being expected by most users most of the time.  Alternatives are encouraged as they do fill a need, but the intended behavior should be the default

hmaarrfk:
Regarding the packages you were concerned with:
* xonsh: [will survive](https://github.com/xonsh/xonsh/blob/b69fb9ab4bf0def1b702db74e016bf43a8478606/xonsh/readline_shell.py#L145)

Nor does it actually use readline 
```python
(xonsh) mark@xps ~ $ import sys; 'readline' in sys.modules                      
False
```

If you really want, I'll make the PR to depend on python-readline, but I really don't think it uses it.

* iPython [doesn't](https://github.com/ipython/ipython/search?q=%22from+readline%22&unscoped_q=%22from+readline%22) use [readline](https://github.com/ipython/ipython/search?q=%22import+readline%22&unscoped_q=%22import+readline%22) (see this [comment for more explanation](https://github.com/conda-forge/python-feedstock/issues/191#issuecomment-414139672))

* Jupyter console: also confirmed not to load in readline

```python
➤ jupyter console                                                                                  (owl) 
Jupyter console 5.2.0

Python 3.6.6 | packaged by conda-forge | (default, Jul 26 2018, 09:53:17) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.5.0 -- An enhanced Interactive Python. Type '?' for help.



In [1]: import sys

In [2]: 'readline' in sys.modules
Out[2]: False
```

Python allows you to **choose** how you to package the distribution. By **choosing** to include the readline module, `conda-forge` is **choosing** to distribute something that has:
  * been linked to GPL'ed code (enforceable I don't know)
  * included a header that claims to be protected by GPL (enforceable I don't know)

Readline offers marginal gains for the python basic interpreter. Having this link in the basic python repository increases the risk of getting a GPL violation.

I hope that the above checks on the imported modules helps reduce your fears of those 3 core packages breaking. For the record, in the original issue I raised, I was also concerned when I first brought up the issue. [Digging deeper](https://github.com/conda-forge/python-feedstock/issues/191) into it, it seems like `readline` isn't so important to how python functions.

hmaarrfk:
(Also xonsh seems really cool!)

hmaarrfk:
How was I able to use conch after deleting readline from my environment?

Clearly it tries to use it from here 

https://github.com/xonsh/xonsh/blob/b69fb9ab4bf0def1b702db74e016bf43a8478606/xonsh/readline_shell.py

mingwandroid:
import readline should still work here no matter what we do! There's readline.py which is used by the Windows implementation and editline of course. Just dropping the module is not ok.

hmaarrfk:
i see

PR #200, opened 2018-09-27
lto-partition fix, again

scopatz:
Here is another fix, with a test

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -ex
 
 # The LTO/PGO information was sourced from @pitrou and the Debian rules file in:
 # http://http.debian.net/debian/pool/main/p/python3.6/python3.6_3.6.2-2.debian.tar.xz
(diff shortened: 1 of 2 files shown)


recipe/build.sh
```diff
@@ -385,3 +385,12 @@ if [[ -n ${HOST} ]]; then
     #                                 -e "s|${PKG_NAME}-${PKG_VERSION}[^ ]*|\${PKG_NAME}-\${PKG_VERSION}|g" > "${RECIPE_DIR}"/sysconfigdata/${our_compilers_name}
     popd
 fi
+
+if [[ ${_OPTIMIZED} == yes && ${CC} =~ .*gcc.* && ${c_compiler} =~ .*toolchain.* ]]; then
+    # On the old toolchain compilers, -flto-partion=none is being replaced
+    # with -partition=none. This needs to be replaced back. Only happens with gcc
+    pushd $PREFIX/
+    sed -i "s/ -partition=none/ -flto-partition=none/g" lib/* lib/*/* lib/*/*/*
```
tkelman: this won't necessarily be gnu sed on mac will it?

scopatz: This shouldn't run on mac since the if statement is for gcc and toolchain

scopatz:
@conda-forge/python this should work now!

PR #207, opened 2018-10-09
Fix find_library when on linux when no objdump present

mingwandroid:
Fixes the objdump issue (that a bug in the previous version of this patch caused, really).


mingwandroid:
Pinging @conda-forge/core

PR #188, opened 2018-07-25
update to 3.6.6

ocefpaf:
@conda-forge-admin, please rerender

PR #212, opened 2018-10-10
Build Python 3.6.6 using separate patches and build script for the compiler variants 

jjhelmus:
This splits the `toolchain` and `gcc/clang` builds with a separate set of patches and build script used for each.  The `toolchain` build closes matches the build as of PR #188.  The `gcc/clang` builds mirror AnacondaRecipes.  

I think this will be a more maintainable setup for this recipe until the compiler migration is complete at which time the `toolchain` patches and build script can be removed.

The downside of this is that the `toolchain` python package no longer has PGO .


jjhelmus:
I'd appreciate a review on this approach (perhaps @ocefpaf or @mingwandroid ).  It is a big change but seem to me to be a safe approach to insure a reliable `toolchain` based build while the compiler migration is in progress.

mariusvniekerk:
Do we need a similar one for py37?

ocefpaf:
> I'd appreciate a review on this approach (perhaps @ocefpaf or @mingwandroid ).

IMO everything looks OK but I'm really not the person to review this :smile:

I trust @mingwandroid and @mariusvniekerk expertise there to merge this one. Thanks @jjhelmus for putting this together!

mingwandroid:
All checks passed, and more tests added, let's roll. Thanks @jjhelmus for taking up the baton.

mingwandroid:
(btw I have no merge rights for this feedstock)

ocefpaf:
> (btw I have no merge rights for this feedstock)

You should though :-/

See https://github.com/conda-forge/python-feedstock/pull/212/files#diff-e178b687b10a71a3348107ae3154e44cR160

ocefpaf:
@mariusvniekerk we are building the old toolchain on CircleCI here too. What was the final decision on that? Or do we expected the lib mismatch to re-appear?

jjhelmus:
> Do we need a similar one for py37?

I think so, I'll check and put together a PR if it is needed this weekend.

PR #215, opened 2018-10-11
Skip osx toolchain

ocefpaf:
I had to pull the OS X packages generated from #212 b/c we are now building OS X on CircleCI instead of Travis-CI and that breaks backward compatibility.

Because the old Python package created with Travis-CI is OK let's just skip it here to avoid breaking the ecosystem every time we need a rebuild.


mariusvniekerk:
LGTM

jjhelmus:
Does this means that if a bug is found in the `toolchain` build for osx we have no method of releasing a fixed version?

ocefpaf:
> Does this means that if a bug is found in the toolchain build for osx we have no method of releasing a fixed version?

Yep. Exactly. Unless someone with more time create a branch and revert the patches that prevent the build from happening on Travis-CI :smile:

Paraphrasing myself here when @mariusvniekerk and I decided to do this: keep the toolchain on life support sucks :stuck_out_tongue_winking_eye:

jakirkham:
That or we can create branches for the `toolchain` build starting from when it last worked.

Issue #201, opened 2018-09-27
Python37 on osx with toolchain broken


mariusvniekerk:
Presently (2018-09-27) for osx py37 with toolchain builds a python that does not work on the very old version of osx that we use on travisCI.

https://anaconda.org/conda-forge/python/3.7.0/download/osx-64/python-3.7.0-h46c1a51_3.tar.bz2

See https://travis-ci.org/conda-forge/randomcolor-feedstock/jobs/434224084

```
dyld: lazy symbol binding failed: Symbol not found: _getentropy
```

This is not a problem that exists when using the Anaconda compilers.

For now we have marked these artifacts as broken on anaconda.org

jakirkham:
Guess the Travis CI log was replaced when restarting unfortunately. That said, looks like Homebrew encountered the same issue ( https://github.com/Homebrew/homebrew-core/issues/4809 ). Maybe we can learn something from their experience.

jakirkham:
Should add some related issues seem to suggest that `MACOSX_DEPLOYMENT_TARGET` is not getting set (properly) or is not recognized by `make`.

jakirkham:
[This issue]( https://gitlab.com/gnutls/gnutls/issues/142 ) over at GnuTLS has a good overview of the problem with some suggested workarounds.

carlodri:
am I wrong or this happens also on python 27?: https://travis-ci.org/conda-forge/staged-recipes/jobs/434497389

from this MR: https://github.com/conda-forge/staged-recipes/pull/6689

mariusvniekerk:
Checked on Sunday and it looked okay

jakirkham:
Right, I think there was some packages published a day or two ago. Hence the question.

jakirkham:
PR ( https://github.com/conda-forge/toolchain-feedstock/pull/46 ) looks like a potential fix.

jakirkham:
Have added PRs ( https://github.com/conda-forge/python-feedstock/pull/217 ), ( https://github.com/conda-forge/python-feedstock/pull/218 ), and ( https://github.com/conda-forge/python-feedstock/pull/219 ) to rebuild the supported Python versions with the legacy compilers on macOS now that PR ( https://github.com/conda-forge/toolchain-feedstock/pull/46 ) is in. We should probably try one of them and make sure it works and if that goes well try the rest.

[...]

PR #217, opened 2018-10-13
Rebuild with toolchain 2.2.2

jakirkham:
There was an issue with the `toolchain`; where, on a newer macOS version, it may not pickup the right SDK. For example, the macOS 10.9 SDK was not being picked up on CircleCI, which runs on a newer version of macOS. Turns out that setting `MACOSX_DEPLOYMENT_TARGET` is not sufficient in these cases and one needs to specify the macOS SDK location as well.

As of `toolchain` 2.2.2, the lazy symbol binding issue has been fixed thanks to the addition of `-isysroot` compiler flag, which specifies the location of the macos SDK we intend to use. Combined with the facts that we specify what minimum macOS version we target and we provide the macOS SDK, this is enough for the build to use that macOS SDK and not pick up these symbols, which do not appear until later macOS versions.

---

Checklist
* [x] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

---

Fixes https://github.com/conda-forge/python-feedstock/issues/201

Related to https://github.com/conda-forge/toolchain-feedstock/pull/46

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,7 +9,7 @@ environment:
     secure: tumuXLL8PU75WMnRDemRy02ruEq2RpNxeK3dz0MjFssnosPm2v4EFjfNB4PTotA1
 
   matrix:
-    - CONFIG: win_c_compilervs2015cxx_compilervs2015vc14
+    - CONFIG: win_c_compilervs2015cxx_compilervs2015target_platformwin-64vc14
       CONDA_INSTALL_LOCN: C:\Miniconda36-x64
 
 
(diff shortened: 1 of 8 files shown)


jakirkham:
Friendly nudge 😉

scopatz:
Sorry for the delay @jakirkham!

jakirkham:
No worries, @scopatz. Thanks for taking a look.

Would you be able to look at PRs ( https://github.com/conda-forge/python-feedstock/pull/218 ) and ( https://github.com/conda-forge/python-feedstock/pull/219 )? They are the same, but are for Python 3.6 and 2.7 instead.

PR #268, opened 2019-08-07
3.8

mingwandroid:
Checklist


recipe/meta.yaml
```diff
@@ -1,40 +1,36 @@
...
+{% set version = "3.8.0b3" %}
 {% set linkage_nature = os.environ.get('PY_INTERP_LINKAGE_NATURE', '') %}
 {% if linkage_nature != '' %}
   {% set linkage_nature = "_" ~ linkage_nature %}
 {% endif %}
 
 package:
   name: python
-  version: {{ version }}
+  version: {{ base_version }}
```
jjhelmus: I am opposed to faking the version here.

recipe/meta.yaml
```diff
@@ -63,62 +59,75 @@ source:
     folder: externals/nasm-2.11.06                                                   # [win]
     sha256: de3c87b26a80e789986d8e6950c6304175d3829afe9c6c7211eb7257266ab0ac         # [win]
 
+
 build:
-  number: 1
+  number: 0
+  add_pip_as_python_dependency: false
```
jjhelmus: disable_pip?

recipe/meta.yaml
```diff
@@ -144,4 +153,3 @@ extra:
     - msarahan
     - pelson
     - ocefpaf
-    - scopatz
```
jjhelmus: should be retained

recipe/sysconfigdata/_sysconfigdata_aarch64_conda_cos7_linux_gnu.py
```diff
@@ -1,805 +0,0 @@
...
- 'WITH_PYMALLOC': 1,
- 'WITH_VALGRIND': 0,
- 'X87_DOUBLE_ROUNDING': 0,
- 'XMLLIBSUBDIRS': 'xml xml/dom xml/etree xml/parsers xml/sax',
- 'abs_builddir': '/opt/conda/conda-bld/python_1512347718438/work/build-static',
- 'abs_srcdir': '/opt/conda/conda-bld/python_1512347718438/work',
- 'datarootdir': '/opt/anaconda1anaconda2anaconda3/share',
- 'exec_prefix': '/opt/anaconda1anaconda2anaconda3',
- 'prefix': '/opt/anaconda1anaconda2anaconda3',
- 'srcdir': '/opt/conda/conda-bld/python_1512347718438/work'}
```
jjhelmus: Needs to be retained

recipe/sysconfigdata/default/_sysconfigdata_linux.py
```diff
@@ -684,9 +684,24 @@
...
+if GCC:
+    with open(devnull, 'w') as err:
+        GCC_VERSION = check_output([GCC, '--version'], stderr=err).decode('utf-8').splitlines()[0].split(' ')[-1]
+    print(GCC_VERSION)
+    for btv in ('BLDSHARED', 'CC', 'CXX', 'LDCXXSHARED',
+                'LDSHARED', 'LINKCC', 'MAINCC'):
+        if btv in build_time_vars:
+            build_time_vars[btv] = build_time_vars[btv]. \
+                replace('-B /opt/anaconda1anaconda2anaconda3/compiler_compat ',
+                        '')
```
jjhelmus: Should this be added to the other linux platforms?

recipe/meta.yaml
```diff
@@ -63,62 +59,75 @@ source:
...
+   # # conda-build bug?
+   #  - ncurses >=6.1,<7.0a0               # [unix]
 
 test:
+  # Because we need to use xvfb-run on linux, from a CDT package.
+  requires:  # [linux]
+    # We are running centos6 software here (on whatever distro you are on)
+    # .. so it best be compatible (you should use centos6 if possible).
+    - {{ cdt('xorg-x11-server-xvfb') }}  # [linux]
+    - {{ cdt('xorg-x11-xauth') }}  # [linux]
```
jjhelmus: These CDT do not exist yet.  They will need to be created.

recipe/build.sh
```diff
@@ -27,20 +95,14 @@ if [[ ${PY_INTERP_LINKAGE_NATURE} == shared ]]; then
   _ENABLE_SHARED=--enable-shared
 fi
 
-# For debugging builds, set this to 0 to disable profile-guided optimization
+# For debugging builds, set this to no to disable profile-guided optimization
```
jjhelmus: Is this still true or does QUICK_BUILD take precedence

recipe/meta.yaml
```diff
@@ -63,62 +59,75 @@ source:
...
-  number: 1
+  number: 0
+  add_pip_as_python_dependency: false
   # Windows has issues updating python if conda is using files itself.
   # Copy rather than link.
   no_link:
     - DLLs/_ctypes.pyd  # [win]
-    - bin/python3.7  # [not win]
-  script_env:
-    - python_branding
```
jjhelmus: I think this is still needed

recipe/bld.bat
```diff
@@ -1,10 +1,10 @@
 setlocal EnableDelayedExpansion
 
-:: brand Python with conda-forge startup message
-%SYS_PYTHON% %RECIPE_DIR%\brand_python.py
-if errorlevel 1 exit 1
+REM brand Python with conda-forge startup message
+REM %SYS_PYTHON% %RECIPE_DIR%\brand_python.py
+REM if errorlevel 1 exit 1
```
jjhelmus: This is needed for conda-forge, unless a patch added similar messaging.

jjhelmus:
The CI is failing because the new CDTs do not exist.  

@mingwandroid Could you target this against the `pre-3.8` branch rather than master?

recipe/ld.wrapper
```diff
@@ -0,0 +1,34 @@
...
+       [[ ${GCC_SYSV} =~ 7[.].* ]]; then
+      $(dirname "${BASH_SOURCE[0]}")/ld.bin "$@"
+    else
+      "${_GCC}" "$@"
+    fi
+  fi
+}
+
+_COMPILER_COMPAT_LD_ARGS=( "$@" )
+ld_fn "${_COMPILER_COMPAT_LD_ARGS[@]}"
```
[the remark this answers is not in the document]
mingwandroid: Just a left over.

recipe/sysconfigdata/default/_sysconfigdata_linux.py
mingwandroid: This is used on all linux platforms when our compilers are not in use is it not? Should be?!

recipe/meta.yaml
mingwandroid: Just `xorg-x11-xauth`. I lost my phone's SIM unlock code (yeah really) so haven't been able to run `c3i` to build it yet. It's ok, I recovered it now.

recipe/build.sh
mingwandroid: Not sure, will get cleaned up.

recipe/meta.yaml
mingwandroid: But not @3.7 right?

recipe/bld.bat
mingwandroid: Yeah, how can we try to keep our recipes in sync. What do you think about using env vars like I do for static vs dynamic Python in meta.yaml? I'm OK with doing something like that personally.

recipe/meta.yaml
mingwandroid: Indeed, sorry.

mingwandroid:
I can indeed target it there. @jjhelmus, do you want to take this over for now? We need version checks for newer and probably some env var overrides. If you want to address only this part independently of the other changes I'm making (which will get more invasive until I have it cleaned up), then feel free to take what I've done in `recipe/sysconfigdata/default/_sysconfigdata_linux.py` and make a separate PR from it to conda-forge and I'll rebase on top next week.

recipe/meta.yaml
mingwandroid: OK, please advise on better suggestions for betas with version comparison considerations, err, considered (carefullly!).

recipe/bld.bat
mariusvniekerk: Sounds cleaner

mariusvniekerk:
closed by #274

PR #277, opened 2019-10-26
Stop vendoring binutils linker

isuruf:
Checklist
* [ ] Used a fork of the feedstock to propose changes

cc @rgommers, @mingwandroid


rgommers:
Thanks @isuruf!

I can't judge if this is the right fix, but assuming it is it will definitely make life on linux-with-modern-gcc easier.

isuruf:
@rgommers, we are still using the linker from conda package, but it's a separate package and will be kept up-to-date.

rgommers:
Ah, in that case - never mind, that `ld` I still have to delete manually on every env creation.

isuruf:
@rgommers, true, but if we kept it up-to-date, would you have to? (there are cases when you'll be using an unreleased binutils and it'll not work in that case)

rgommers:
> @rgommers, true, but if we kept it up-to-date, would you have to?

yep, I usually default to `defaults`, the `ld` from that channel doesn't support gcc 9

isuruf:
What's your binutils version? `ld --version`

rgommers:
```
$ ld --version
GNU ld (GNU Binutils) 2.32
```

isuruf:
Thanks. @conda-forge, we have 2.32 and 2.33.1 is on the way. I guess @mingwandroid or @jjhelmus can build the binutils package using conda-forge recipes and upload to `defaults`.

jjhelmus:
I'm in favor of this idea but it does not appear as if `ld_impl_linux-64` packages exist.

isuruf:
@jjhelmus, I removed them until https://github.com/conda-forge/binutils-feedstock/pull/18 is fixed. I could get linux-64 to pass, but couldn't get ppc64le nor aarch64 to pass.

jakirkham:
Maybe we could add a selector to get the architectures we can cover?

isuruf:
Ready now

scopatz:
LGTM!

jjhelmus:
Thanks @isuruf, this is great.

PR #279, opened 2019-11-02
Windows rework patches.

mariusvniekerk:
This is an attempt to restructure the windows DLL loading patches


mingwandroid:
Personally, I prefer `0020` to `0019` and think we should continue the work in `0020` instead.

mingwandroid:
To be clear, dropping respect for `PATH` will break things, and adding back just conda path entries will not fix everything. Now this is an upstream decision, but it's also a potentially very destructive one.

mariusvniekerk:
Yeah, but it seems like we're fairly forced by cpython's changes here

mariusvniekerk:
To be precise, all that this change does is to move the change by @jjhelmus to run before the LoadLibraryEx call, rather than after it

ocefpaf:
Tested it locally against some packages that are broken at the moment and it works. Let's get this merged so se can move forward.

jjhelmus:
Moving the `AddDllDirectory` calls before `LoadLibraryEx` make sense.  `LoadLibraryEx` performs the loading of the library, if the conda paths have not been added to the search order then this will fail is the library needs to resolve a library that is only available in one of those paths.  

The `AddDllDirectory` after `LoadLibraryEx` was likely working in some cases because another library was loaded before the one requiring the altered search path.  

I believe this process could be improved by only adding the conda paths once rather than each time a library is loaded.  A global variable to track if these paths have been added should do the trick.

I also agree with @mingwandroid, the logic in the original `0019` patch was helpful for debugging and was more efficient that the current patch.  The two should be reconciled into a single patch.

jakirkham:
Could you please open an issue on that @jjhelmus (if there isn’t one already)?

jjhelmus:
For my own benefit as well as anyone who is interested, Microsoft has good docs on this topic: https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order?redirectedfrom=MSDN

jjhelmus:
> Could you please open an issue on that @jjhelmus (if there isn’t one already)?

I created #290 to track this.

ocefpaf:
Thanks @jjhemuls!

PR #285, opened 2019-11-11
Don't py_compile python2 syntax test files

duncanmmacleod:
This PR adds the 2to3 input test files to the `skip_compile_pyc` list. `py_compile` is [currently broken](https://bugs.python.org/issue38731) meaning that if any module can't be compiled, the whole thing dies, so we need to work around that for now (not compiling these test inputs is the right thing to do anyway, IMO).

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -70,7 +70,7 @@ source:
 
 
 build:
-  number: 3
+  number: 4
   # Windows has issues updating python if conda is using files itself.
   # Copy rather than link.
   no_link:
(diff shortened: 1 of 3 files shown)


scopatz:
This looks good to me! Any other feedback?

recipe/meta.yaml
```diff
@@ -70,14 +70,16 @@ source:
...
+  number: 4
   # Windows has issues updating python if conda is using files itself.
   # Copy rather than link.
   no_link:
     - DLLs/_ctypes.pyd  # [win]
   script_env:
     - python_branding
   skip: true  # [win and vc != '14.1']
+  skip_compile_pyc:
+    - lib/python3.*/lib2to3/tests/data/*.py  # [py>=30]
```
jjhelmus: There is not need for the selector here and I'm not entirely sure how it resolves.

hmaarrfk:
Is this still relevant in late 2022?

PR #286, opened 2019-11-11
Patch around bpo-38731

duncanmmacleod:
This PR adds a patch for [bpo-38731](https://bugs.python.org/issue38731) which impacts the files included in this package, and any downstream packages for which `python -m py_compile` might fail.

Checklist

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -16,9 +16,7 @@ steps:
     BINSTAR_TOKEN: 
       from_secret: BINSTAR_TOKEN
   commands:
-    - pwd
-    - export FEEDSTOCK_ROOT="${CI_WORKSPACE:-$(pwd)}"
-    - echo "$FEEDSTOCK_ROOT" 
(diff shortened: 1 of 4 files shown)


beckermr:
LGTM! but my opinion doesn't matter so much here

duncanmmacleod:
This probably supersedes #285, but IMO they are both worthy of consideration.

beckermr:
To add some details here, `conda build` relies on `py_compile` and its proper functioning during errors. This bug is thus not simply a bug in `python` but it also causes downstream issues with `conda build` making `.pyc` files. For this reason, I think it should be patched here in the feedstock until an upstream fix is pushed. 

A patch was put in for a bug in how `python` used bin utils so that downstream builds w/ `conda build` work. The situation here is the same.

duncanmmacleod:
Ping @conda-forge/python.

PR #291, opened 2019-12-13
Testing some Windows python 3.8 issue.

mingwandroid:


mingwandroid:
This is not yet ready for review. I'm leveraging CF CI to investigate a WIndows Python 3.8 bug mostly (but I also want to resync AnacondaRecipes/python-feedstock with conda-forge/python-feedstock where possible).

msarahan:

recipe/meta.yaml
```diff
@@ -65,26 +68,33 @@ source:
...
   script_env:
     - python_branding
+#   - PY_INTERP_LINKAGE_NATURE
   skip: true  # [win and vc != '14.1']
   # We delete the shared libraries.
   ignore_run_exports:   # [unix]
     - bzip2             # [unix]
+  detect_binary_files_with_prefix: False
+  # match python.org compiler standard
+  skip: True            # [win and vc != 14]
```
msarahan: Should be less than 14, not != 14
```suggestion
  skip: True            # [win and vc < 14]
```

recipe/meta.yaml
```diff
@@ -65,26 +68,33 @@ source:
...
   no_link:
     - DLLs/_ctypes.pyd  # [win]
   script_env:
     - python_branding
+#   - PY_INTERP_LINKAGE_NATURE
   skip: true  # [win and vc != '14.1']
   # We delete the shared libraries.
   ignore_run_exports:   # [unix]
     - bzip2             # [unix]
+  detect_binary_files_with_prefix: False
```
jjhelmus: Disabling binary file detection was reported to cause issues when embedding Python, ContinuumIO/anaconda-issues#11439.  I was planning or removing this from the AnacondaRecipes version.

mingwandroid:
Thanks guys!

recipe/meta.yaml
mingwandroid: I want to prefix replace only where necessary, particularly for python as it's frequently used and has a lot of binaries that contain the prefix (mostly harmlessly). If I can't get it to work I'll remove this bit.

recipe/meta.yaml
mingwandroid: Whoops!

PR #292, opened 2019-12-15
Master 3.7 cf

mingwandroid:


mariusvniekerk:
Linux seems to be missing some kind of patch so it aint finding libs.
For windows we should copy over the compiler overrides in 3.8 to deal with the msbuild version requirement change.

mingwandroid:
Hey @mariusvniekerk,the failure is somehow caused by enabling PGO. This is a bug that we thought was due to disabling prefix replacement but there's something else wrong too!

If I set _OPTIMIZED=no then the build works fine.

mingwandroid:
OK I got to the bottom of the problem with embedding python and added a test for embedding with both the static and shared libpythons. Note, for the static case (but probably not for the shared case) it is essential that the exe be located in the correct relative path to the shared libpython (or the other way around really!)

Hopefully this PR will be ready soon, then I'll port the improvements over to python 3.8 and merge back to AnacondaRecipes.

Cheers.

recipe/conda_build_config.yaml
```diff
@@ -0,0 +1,8 @@
+python:                        # [win]
+  - 3.8                        # [win]
```
jjhelmus: This should be 3.7 or removed.

mingwandroid: 3.7 yeah, thanks

mariusvniekerk: may as well be removed

mariusvniekerk:
ooh windows is green

mingwandroid:
I need to put PGO back in for Windows yet and figure out what's going on with my attempt to limit the prefix replacements on Linux. Getting there. 3.8 started too.

mingwandroid:
Windows PGO fails:

LINK : fatal error LNK1104: cannot open file 'pgort.lib' [%SRC_DIR%\PCbuild\pythoncore.vcxproj]

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=105253&view=logs&j=21f341a4-37ca-57ed-1287-2998d263b060&t=417b2021-e743-51fa-ff2a-b94cccea1b49&l=1033

I will have to disable this again I think.

mariusvniekerk:

mingwandroid:
Drone failing here?

mariusvniekerk:
Awesome.  Anything else you want to add before we merge?

mingwandroid:
I think this is good to go now. There'll hopefully be more improvements (size, PGO on Windows) soon! PGO works locally for me with VS2019 FWIW.

mariusvniekerk:
Yeah we need to just install the right additional vs2017 packages to get pgo going

jakirkham:
Do we know what those are or where to get them?

mariusvniekerk:
@jakirkham I've got a PowerShell snippet I'm testing out that might work

mariusvniekerk:
Something like this seems to work https://github.com/conda-forge/_azure_pipelines_test/blob/master/azure-pipelines.yml#L90-L131
Takes about 10 minutes or so.

mariusvniekerk:
Tried to run the install for what we thought was the required package.

See the following log for the details:
https://dev.azure.com/conda-forge/84710dde-1620-425b-80d0-4cf5baca359d/_apis/build/builds/105965/logs/56

jakirkham:
Does that mean PGO is already working on Windows?

mingwandroid:
No, but AFAIK we've never had PGO on windows. At least since the move to AP no python build have had PGO enabled. This isn't a regression, I'm just making people aware.

Once we figure out how to install the package, comment out set PGO= in bld.bat

If you do that before then then you'll get a build failure.

PR #295, opened 2019-12-20
Update to 3.8.1 and add some tests

mingwandroid:
I don't expect this to work well for a while, but I want to get it up on CI to see where things stand.


mingwandroid:
I am baffled as to why patching fails on Windows. The patches have the correct CRLF status (patch lines use LF and the patched Windows files - .bat and .vcxproj mostly) use CRLF, I've added m2-patch as a build dep, .gitignore tells it to treat *.patch as binary, it works fine on the other OSes.

Thoughts welcome! (I wish github had an equivalent of show whitespace or cat -e at times like this).

I will enable PGO next anyway and hope that Windows magically fixes itself overnight.

mariusvniekerk:
If i remember correctly the pgo libraries aren't installed on these machines? cc @jjhelmus

mingwandroid:
The tests added are important but some of them have python deps which will create trouble of a cyclic nature when we update to the next minor release. noarch: python ones will help, failing that we'll need to add jinja2 to skip them on these initial buildouts (3.9 will be the next one).

mingwandroid:
.. and Travis failing here.

mingwandroid:
This is ready for review. Would you prefer I squashed all the commits though?

mariusvniekerk:
Squashing would be nice.  Same for 3.7 i think

recipe/build.sh
```diff
@@ -288,20 +299,18 @@ if [[ ${_OPTIMIZED} == yes ]]; then
...
-else
-  make -C ${_buildd_shared} install
-  declare -a _FLAGS_REPLACE
 fi
+# Install the shared library (for people who embed Python only, e.g. GDB).
+# Linking module extensions to this on Linux is redundant (but harmless).
+# Linking module extensions to this on Darwin is harmful (multiply defined symbols).
+cp -pf ${_buildd_shared}/libpython*${SHLIB_EXT}* ${PREFIX}/lib/
+if [[ ${target_platform} =~ .*linux.* ]]; then
+  ln -sf ${PREFIX}/lib/libpython${VERABI}${SHLIB_EXT}.1.0 ${PREFIX}/lib/libpython${VERABI}${SHLIB_EXT}
```
mariusvniekerk: This seems to be missing the `.1` symlink?  Does this matter much?

recipe/conda_build_config.yaml
```diff
@@ -1,8 +1,14 @@
...
   - vs2017                     # [win]
-vc:                            # [win]   
-  - 14.1                       # [win] 
+vc:                            # [win]
+  - 14.1                       # [win]
+fortran_compiler_version:      # [win]
+  - 2019.0.0                   # [win]
+nodejs:                        # [win]
+  - 8.12                       # [win]
+geos:                          # [win]
```
mariusvniekerk: Why this one?

mariusvniekerk: @mingwandroid can you answer those two comments please?  Just want to be sure about them before merging.  Thanks

mingwandroid:
@mariusvniekerk I think I am targeting the wrong branch here, should it be master instead?

CJ-Wright:

jakirkham:
Yeah I think the bot is no longer able to handle re-rendering PRs from orgs correctly. Have the same issue.

jschueller:
can you rebase to master ? it would be helpful to only see the relevant commits

PR #296, opened 2019-12-22
Add libbluetooth dev deps to add python support to bluetooth sockets

sergioisidoro:
### ⚠️ Disclaimer: I'm new around here ⚠️

I was playing around with python and BLE, and stumbled upon this old issue. 
I decided to give it a try

https://github.com/ContinuumIO/anaconda-issues/issues/1035

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -95,6 +95,7 @@ requirements:
     - llvm-tools                         # [osx]
     - make                               # [not win]
     - pkg-config                         # [not win]
+    - libbluetooth-dev                   # [not win]
   host:
     - bzip2             # [not win]
     - ld_impl_{{ target_platform }}        # [linux and not aarch64]
@@ -108,6 +109,7 @@ requirements:
     - ncurses           # [not win]
     - libffi            # [not win]
     - vc                # [win]
+    - libbluetooth-dev  # [not win]
   run:
     - ld_impl_{{ target_platform }}        # [linux and not aarch64]
     - binutils_impl_{{ target_platform }}  # [aarch64]
@@ -119,6 +121,7 @@ requirements:
     - tk                # [not win]
     - ncurses           # [not win]
     - libffi            # [not win]
+    - libbluetooth-dev  # [not win]
 
 test:
   requires:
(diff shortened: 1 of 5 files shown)


sergioisidoro:
Ok, this is clearly a bit out of my reach as a beginner right now :) 

So far, as I could gather the deps are`libbluetooth-dev` in debian, `bluez-dev` in alpine, and I have very little knowledge about windowns platform.

If anyone would point me in the right direction I might still try, but I'm a bit lost right now.

recipe/meta.yaml
```diff
@@ -92,6 +92,7 @@ requirements:
     - {{ compiler('cxx') }}
     - {{ cdt('xorg-x11-proto-devel') }}  # [linux]
     - {{ cdt('libx11-devel') }}          # [linux]
+    - {{ cdt('libbluetooth-dev') }}      # [not win]
```
hmaarrfk:
```suggestion
    - {{ cdt('libbluetooth-dev') }}      # [linux]
```

hmaarrfk:
ohhh, this one is hard. one issue I see happening is that you will have to  think about what works for `cos6` and that might not be what works for `cos7`, which means that you might not be able to install your `cos6` package on ubuntu.

You might need to add things like cdts for the things you need. Check staged recipes for how to do that.

Do you need to change anything for OSX?

scopatz:
Yes, this need cos7 to work, which we don't have yet

recipe/meta.yaml
sergioisidoro: This is something I found out later. it's `libbluetooth-dev` for debian / ubuntu, but `bluez-dev` for things like alpine distributions. So I have no idea on how to specify things on a per distro.

sergioisidoro:
> Do you need to change anything for OSX?

I'm not really sure... As far as I could gather, python needs to be built with `bluetooth.h` so that `AF_BLUETOOTH` doesn't at least throw an undefined error. Apple seems to have some remainders of documentation for that header, and things seemed to work on my machine, if I recall correctly 

https://developer.apple.com/documentation/kernel/bluetooth_h?language=objc

hmaarrfk:
This is close to being possible. I think there is going to be a cos6->cos7 migration soon, do you want to open an issue about this?

PR #310, opened 2020-02-08
noarch run_exports

isuruf:
Checklist
* [ ] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -78,7 +78,7 @@ source:
 
 
 build:
-  number: 3
+  number: 4
   # Windows has issues updating python if conda is using files itself.
   # Copy rather than link.
   no_link:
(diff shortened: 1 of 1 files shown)


jjhelmus:
What will versions of conda-build prior to 3.18.12 do with this `run_export`.  My reading of conda/conda-build#3868 is that the `weak` run_export will be applied in this case.  

Not a reason to hold this up, just something to keep in mind. 

It might be good to ensure that conda-build >=3.18.12 is used when building `noarch` packages if not all packages.

isuruf:
@jjhelmus, I agree. I'll add a condition to conda-forge-ci-setup.

I first thought of adding a check to conda-build to error out if an unknown `run_exports` type is used by a package, but that's not going to work with older conda-build releases, so I didn't. If you think it'll be useful for a new type of run_exports, I can send a PR.

PR #314, opened 2020-02-25
Add _dbg support for Windows, add more in-build testing (needs a new …

mingwandroid:
…conda I fear due to using conda run there)

Checklist
* [ ] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)


recipe/meta.yaml
```diff
@@ -78,7 +80,7 @@ source:
 
 
 build:
-  number: 3
+  number: 8
```
scopatz: That is a pretty big build number jump

recipe/bld.bat
```diff
@@ -166,30 +207,39 @@ if errorlevel 1 exit 1
...
+echo "Testing print('Hello') prints Hello"
+conda run -p %PREFIX% python -c "print('Hello')" 2>&1 | findstr /r /c:"Hello"
 if %errorlevel% neq 0 exit /b 1
 
-echo "Testing import of os (no DLL needed) does not print: The specified module could not be found"
-%CONDA_EXE% run -p %PREFIX% cd %PREFIX% & %PREFIX%\python.exe -v -c "import os" 2>&1
-%CONDA_EXE% run -p %PREFIX% cd %PREFIX% & %PREFIX%\python.exe -v -c "import os" 2>&1 | findstr /r /c:"The specified module could not be found"
+echo "Testing import of os does not print The specified module could not be found"
+conda run -p %PREFIX% python -v -c "import os" 2>&1
+conda run -p %PREFIX% python -v -c "import os" 2>&1 | findstr /r /c:"The specified module could not be found"
```
scopatz: I am not sure I understand why the variable names are replaced

mingwandroid: Because %CONDA_EXE% is not what should be run by anyone, ever, including scripts.

conda is, on Windows, only ever allowed to be a shell function or script (a batch file or a powershell functions). It is critial for its correct behaviour that this is the case. I could go into the details but it's down to the fact that the scripts/batch files correctly prefix PATH with the 6 entries needed for the correct execution of the real conda.exe before calling it.

recipe/meta.yaml
mingwandroid: This is a sync with AR and therefore I prefer to sync build numbers too. It means I know whether packages are largely equivalent.

scopatz: Yeah that's fine!

PR #331, opened 2020-03-20
Fix brand regex

SylvainCorlay:
Fixes #194 

**I would really advocate from completely removing the `brand_python.py` though. I don't think that modifying `_sys_version` is a clean way of changing the banner.**

Checklist

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -10,8 +10,12 @@ jobs:
   strategy:
     maxParallel: 8
     matrix:
-      linux_target_platformlinux-64:
-        CONFIG: linux_target_platformlinux-64
+      linux_python3.6.____cpythontarget_platformlinux-64:
+        CONFIG: linux_python3.6.____cpythontarget_platformlinux-64
+        UPLOAD_PACKAGES: True
+        DOCKER_IMAGE: condaforge/linux-anvil-comp7
+      linux_python3.7.____cpythontarget_platformlinux-64:
+        CONFIG: linux_python3.7.____cpythontarget_platformlinux-64
         UPLOAD_PACKAGES: True
         DOCKER_IMAGE: condaforge/linux-anvil-comp7
   steps:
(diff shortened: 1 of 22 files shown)


SylvainCorlay:
This windows build failure appears to be unrelated to that change:

```
ERROR conda.core.link:_execute(568): An error occurred while installing package 'conda-forge::pkginfo-1.5.0.1-py_0'.
CondaError: Cannot link a source that does not exist. C:\Miniconda\Scripts\conda.exe
```

isuruf:
@conda-forge-admin, rerender

SylvainCorlay:
Even if this works, I think we should consider completely removing that patch.

jjhelmus:
I'm fine removing this patch entirely.  It was in conda-forge before I started working on the python recipe so I kept it.

mingwandroid:
What about using the anaconda branding mechanism instead? We change the c code. I can template the patch and apply the correct replacement at build time if that's acceptable?

Branding is kinda useful if you change things significantly and I think, for better or worse we do that. But I don't mind if we remove it entirely for conda-forge either really.

jakirkham:
Having some kind of branding is useful I think. Though I don't know if we are discussing removing it or just some minor tweaks at this point.

SylvainCorlay:
> What about using the anaconda branding mechanism instead? We change the c code. I can template the patch and apply the correct replacement at build time if that's acceptable?

I don't know where to find that recipe?

SylvainCorlay:
> Having some kind of branding is useful I think. Though I don't know if we are discussing removing it or just some minor tweaks at this point.

This PR does a minor tweak but I was suggesting we should consider removing it completely.

SylvainCorlay:
For your information, I opened #332 completely removing the patching of `sys.version` - and which I think preferable over this approach.

SylvainCorlay:
Any interest in getting this in?

isuruf:
Can you bump the build number?

SylvainCorlay:
> Can you bump the build number?

Done

SylvainCorlay:
@isuruf  I bumped the build number - and did some more testing locally.

SylvainCorlay:
The rerender seems to be problematic. I don't see why there should be a `linux_python3.7` build of Python 3.8...

isuruf:
One build in linux-64 passing is enough

Issue #330, opened 2020-03-20
Branding patches cause issues with certain wheels


SylvainCorlay:
This recipe patches the CPython source code to include **conda-forge branding** in the interpreter banner, namely `| packaged by conda-forge |`. However, since this banner is validated in `platform.py` (to prevent mismatches between the Python runtime and the python home in use), the `platform.py` is also patched for this validation to pass. Both patches are done in [brand-python.py](https://github.com/conda-forge/python-feedstock/blob/master/recipe/brand_python.py).

Now, I have the following issue with this: we are working on a PyPI wheel for the xeus-python jupyter kernel, for which we have a functional conda package, and which embeds Python. The conda package simply links with libpython.so and all is fine, but for the wheel, since we generally can't rely on Python distributions providing a `libpython.so` shared object, our PyPI manylinux wheel is statically linked with `libpython.a`. This causes the banner to *not* be the one from conda-forge an to cause a mismatch with conda-forge's patched validation regex.

To solve this issue, we would need to either renounce to the branding in the banner, or make the new regex slightly more flexible to also accept the original banner from the unpatched python interpreter.

cc @mingwandroid @jjhelmus @johanmabille.

PR #332, opened 2020-03-22
Remove conda-forge branding

SylvainCorlay:
Alternative to #331. 
I think that this fix is actually preferable than the current branding solution.


SylvainCorlay:
> -1 on this. I know that there are some downstream software which use sys.version to detect a conda installed python.

Patching `sys.version` to add brand the banner is hacky at best (and the fact that platform.py much be patched as well shows that it has side effects at least). Issue #330 shows that has negative consequences. 

I really prefer this approach over making the regex more tolerant.

mingwandroid:
Playing devil's advocate here but given

> our PyPI manylinux wheel is statically linked with libpython.a. This causes the banner to not be the one from conda-forge an to cause a mismatch with conda-forge's patched validation regex.

Isn't this the thing we should fix here first. Then branding or no branding is untangled from your issue.

jakirkham:
Agree with Ray 🙂

SylvainCorlay:
> Playing devil's advocate here but given
>
> > our PyPI manylinux wheel is statically linked with libpython.a. This causes the banner to not be the one from conda-forge an to cause a mismatch with conda-forge's patched validation regex.
> 
> Isn't this the thing we should fix here first. Then branding or no branding is untangled from your issue.

@mingwandroid @jakirkham  Quite the contrary. The conda package for xeus-python is dynamically linked with libpython.so, but in the case of a wheel, some distributions of Python don't come with a shared object - and we must link statically. This is how several other packages operate to distribute PyPI wheels.

My point is that conda-forge's Python brands the Python interpreter's banner by changing the `sys.version` string, which has several side effects as the string is validated at runtime. I don't think this is a good thing to do at all.

mingwandroid:
I must not have made myself clear here.

That branding works with the dylib but not the static archive seems to me a serious bug and entirely orthogonal to whether we should brand or not.

Now if everyone was in agreement to ditch branding then OK, let's do that, delete all the changes and move on.

Bur that's not the situation and if we did do that we'd need a deprecation cycle for it.

I'm happy to look into using the anaconda branding code for conda forge, which I believe works in both cases. In my opinion this is the best path forward.

SylvainCorlay:
> That branding works with the dylib but not the static archive seems to me a serious bug and entirely orthogonal to whether we should brand or not.

Actually, there is no bug, since the manylinux wheel is linked against the `libpython.a` from manylinux and not conda, which is why it does not get the branding.

> we'd need a deprecation cycle for it.

I think we can start with #331 which makes the regex compatible with both the branding and the absence thereof, and discuss the complete removal at greater length.

mingwandroid:
What manylinux libpython.a do you mean? AFAIK this doesn't exist. You are linking against some random libpython the distribution you ran your build upon happened to have installed I guess?

I don't want to spend time taking about random software.

mingwandroid:
And is it true that you are using this random python but with a python syslib from conda-forge?

SylvainCorlay:
> You are linking against some random libpython the distribution you ran your build upon happened to have installed I guess?

No, we are linking against the right libpython.a, but preventing the manylinux docker file from deleting it. This is something several other packages are doing to embed python, such as Panda3D.

PR #345, opened 2020-04-24
update to 3.7.7; apply DllSearchPath patch also on 3.7-branch

h-vetinari:
Checklist

I have [reason](https://github.com/conda-forge/cryptography-feedstock/pull/37#issuecomment-619172197) to believe that this patch is necessary for being able to use `CONDA_DLL_SEARCH_MODIFICATION_ENABLE`, which seems to be needed for https://github.com/conda-forge/cryptography-feedstock/pull/37 according to @mingwandroid.


h-vetinari:

recipe/patches/0019-Add-CondaEcosystemModifyDllSearchPath.patch
```diff
@@ -1,456 +1,447 @@
...
-@@ -1238,7 +1244,7 @@ pymain_header(_PyMain *pymain)
-         return;
-     }
- 
--    fprintf(stderr, "Python %s :: %s on %s\n", Py_GetVersion(), Anaconda_GetVersion(), Py_GetPlatform());
-+    fprintf(stderr, "Python %s :: %s on %s\n%s", Py_GetVersion(), Anaconda_GetVersion(), Py_GetPlatform(), CondaEcosystemGetWarnings());
-     if (!Py_NoSiteFlag) {
-         fprintf(stderr, "%s\n", COPYRIGHT);
-     }
-@@ -3017,9 +3023,388 @@ pymain_init(_PyMain *pymain)
```
h-vetinari, line 50: Need to skip this part because this builds on `0001-Add-Anaconda-Distribution-version-logic.patch`, which is applied neither on [3.7](https://github.com/conda-forge/python-feedstock/blob/3.7/recipe/meta.yaml#L17-L19) nor on [master](https://github.com/conda-forge/python-feedstock/blob/d0ab36792688c2531d6d89e5ad3f201af174094b/recipe/meta.yaml#L20-L22).

h-vetinari:
The patch rebase was done by applying everything on `v3.7.6` and then running `git format-patch`. Unfortunately, git ignores the actual endings of the source files when creating the patches. The way to solve this is (broadly) to convert line endings for patches of CRLF-files to CRLF as well.

I've been a bit more careful and now the patches themselves are LF-only wherever possible (all lines except the actual payload of the diff). There's a funny edge case when the original file has no line ending, see https://github.com/conda-forge/python-feedstock/pull/345/commits/0c87042050074182af22a037d43db395701e8af3

In any case, please checkout the in ⚙️-->"Hide whitespace changes" option in the diff viewer when reviewing the patches.

h-vetinari:
Discounting some errors during the python test suite on OSX/Linux (which seem to be the norm however; see #346), this should be finished. Please see [this comment](https://github.com/conda-forge/python-feedstock/pull/345#issuecomment-619252139) about the patch rebase.

mingwandroid:
Hey @h-vetinari, to fix the CRLF patches I use: https://github.com/AnacondaRecipes/aggregate/blob/master/make-mixed-crlf-patch.py

This commit addresses a lot of issues about how conda-build drives GNU patch, the (currently messy) comments go into some detail about the issues with the old code: https://github.com/mingwandroid/conda-build/commit/73a98649ae4874b73ac0cbcbd4733d5b4f6daa67#diff-d002a600d4545ed9ea05d15187120389

Cheers!

h-vetinari:
@mingwandroid 
If I understand that comment correctly, I have correctly formatted all the patches as "native" patches here.

I'm not sure if your message is meant to make me change something about this PR? If so, what?

@isuruf [mentioned](https://github.com/conda-forge/cryptography-feedstock/pull/37#issuecomment-619296276) that he doesn't think any CF-packages should use `CONDA_DLL_SEARCH_MODIFICATION_ENABLE`, what do you think about the whole thing?

I'm guessing it doesn't hurt to add the *capability* to use it, which patch 0019 here seems to be doing. I have some more [questions](https://github.com/conda-forge/cryptography-feedstock/pull/37#issuecomment-619353671) about that, but I'll leave those for the other PR.

I can also revert https://github.com/conda-forge/python-feedstock/pull/345/commits/9d07cce37f4df83fdae4f29567177b0b10b56709 for the time being, and the rest of this PR would build fine just the same.

mingwandroid:
No, I was just pointing you to a simple script that might help and hoping to make you aware of some other bugs in how patching is performed. There was no suggestion of changing anything.

I am conflicted about CDSME. It does too much, IMHO it should be avoiding System32 but *nothing else*. That is, in fact, why I originally wrote it. So that people with Amplitube installed could use MKL.

Since then we've seen OpenSSL DLLs appearing in C:\Windows\System32, and I *think* that may be MS' doing.

I agree with @isuruf, we shouldn't set it on activate, that should be up to end-users and it's better to set it during builds only and advise people who have trouble that it is an option for them. Cheers.

scopatz:
This looks good to me!

h-vetinari:
> @mingwandroid: No, I was just pointing you to a simple script that might help and hoping to make you aware of some other bugs in how patching is performed. There was no suggestion of changing anything.

Thanks for that. It will hopefully save me some manual CRLF replacement work next time. :)

> @mingwandroid: I agree with @isuruf, we shouldn't set it on activate, that should be up to end-users and it's better to set it during builds only and advise people who have trouble that it is an option for them. Cheers.

But then the following [question](https://github.com/conda-forge/cryptography-feedstock/pull/37#issuecomment-619353671) arises (to me) in the context of https://github.com/conda-forge/cryptography-feedstock/pull/37:

> @h-vetinari: I was also wondering: It's fair enough that vendors shouldn't install into `C:\Windows\System32`, but what if an end user does have those libs there already? Then they'd have a (potentially) vulnerable environment, while conda happily tells them they're on the newest openssl-version.

h-vetinari:
Also, I couldn't get the patch to work on the 3.6-branch, see #347. If you have them running on 3.6 within anaconda, a sync here would be much appreciated.

mingwandroid:
In general, a random openssl in that dir is likely to be incompatible Abi wise with ours so will fail to load.

At the end of the day windows dll loading is really poor and insecure but it's also the only thing we have (until something like midipix is ready for prime time use).

We could add a symbol to our openssl and force linking to that somehow whenever the header is included or something?

If CDSME did only system32 avoidance I would not object to recommending we set it globally or defaulted it to enabled though really. It does address the system32 problem well I think.

h-vetinari:
I answered the `cryptography`-related things in the other [thread](https://github.com/conda-forge/cryptography-feedstock/pull/37#issuecomment-619450544).

As for this PR, I think it should be ready to go as is? Still outstanding what to do on python 3.6 (although we could raise with my proposal from the other thread). If you have a version of that `CONDA_DLL_SEARCH_MODIFICATION_ENABLE`-patch on 3.6, it would be awesome if you could help out with #347. The implementation of pymain changed *a lot* from 3.6 to 3.7, and since building your patch failed on some aspects of that, I gave up.

h-vetinari:
Anything left to do here?

As a reminder:
> @h-vetinari: I can also revert https://github.com/conda-forge/python-feedstock/pull/345/commits/9d07cce37f4df83fdae4f29567177b0b10b56709 for the time being (in case more discussion is necessary), as the rest of this PR would build fine just the same.

h-vetinari:
Ping @conda-forge/python

h-vetinari:
@conda-forge/python 
Anything left to do here?

h-vetinari:
It's dead, Jim

PR #348, opened 2020-05-19
update to 3.8.3

jjhelmus:
Checklist


jjhelmus:

jjhelmus:
I moved the `cpython` text to the beginning of the build string in 162be59.  @kalefranz noticed that there are a few places in conda and other tools that assume that package names end with a build number.  Since the `cpython` here is only intended to inform users and should not be used to select packages the location of the text should not matter.  If matching is needed `cpython*` or `*cpython*` could be used.
@isuruf does this seem reasonable?

isuruf:
Sure. If we are making this change, we should make the change to 3.6, 3.7 and pypy3.6 though.

jakirkham:
It also shows up in the docs. Would we have to fix anything else (like `conda-forge-pinning`, repodata patches, etc.)?

beckermr:
We need to fix

- migrations files
- pypy build strings
- conda-forge pinnings
- admin web services command
- repodata patches

jjhelmus:
Let's merge this, I'll submit PR for the other Python versions.

jjhelmus:
@mbargull mentioned this change in today's conda-forge meeting.  I would be alright if this was reverted.  The build number is at the end of the build string in Python 3.8.x and 3.7.x.  #362 would be a good place to change it for 3.7.8 if there is interest.

PR #362, opened 2020-07-22
3.7.8

jjhelmus:
Checklist


jjhelmus:

jjhelmus:

jjhelmus:
This is passing CI but I would like to resolve the question no where `cpython` should appear in the build string (at the end or beginning) prior to merging this.  @isuru and @mbargull seem most interested in this.

mbargull:
@isuruf (missing "f" above ;).)

Yes, please return to the `h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}_cpython` scheme -- then we don't have to change anything downstream in `conda-forge-pinning` etc. :).

----
I also did a quick review: Apart from the build string, this looks alright to me (not know if anything else changed in CPython).
But man, viewing `diff`s of `diff`s really does not have the nicest highlighting on GitHub!

jjhelmus:
I've moved `cpython` to the end of the build string.

Usually the patch diffs are cleaner and easier to review, for example in #358.  I think the refresh here included a change in line endings which causes the who file to get highlighted.

mbargull:
> I think the refresh here included a change in line endings which causes the who file to get highlighted.

Ah right. Would be good if it highlighted carriage returns (at least when they are added/removed)...
When Ray updates things again, we'd have the same situation once more in the opposite direction then, I guess.

PR #366, opened 2020-07-27
ENH add new sysroot name for 3.6

beckermr:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #367, opened 2020-07-28
CFEP-18: Remove static libraries from main build

### LINKED ISSUE #260: Split out static library
The libpython static library is quite large

```
$ du -h ./lib/libpython3.7m.a
 16M    ./lib/libpython3.7m.a
```

and won't be used by most users. It would be useful to split this out into a separate package if possible, to reduce installation size (important for docker images).

xhochy:
Fixes #260 

Checklist


xhochy:
Also seeing the Linux error in https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=189743&view=logs&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=841356e0-85bb-57d8-dbbc-852e683d1642&l=791 in #366 which is not using outputs, so this might be unrelated.

beckermr:
These lines are very odd. The old sysroot paths should not be in this build.

```
checking build system type... x86_64-conda_cos6-linux-gnu
checking host system type... x86_64-conda_cos6-linux-gnu
```

beckermr:
@xhochy This failure is due to a bug in the compiler activation for the new sysroot. It should be fixed within a few hours or so.

mingwandroid:
Great stuff!

xhochy:
The Linux issue is the same as in #368 , continuing with Windows here then.

xhochy:
Verified that the Windows build currently works with the same script, just without using `outputs:`, so this is probably an `outputs` problem / different behaviour.

xhochy:
@conda-forge/python Any idea why this fails on Windows?

xhochy:
Thanks @mbargull!

xhochy:
@conda-forge/python This is ready for review/merge.

recipe/meta.yaml
```diff
@@ -116,83 +116,119 @@ build:
...
+        -   pushd prefix-replacement  # [unix]
+        -     bash build-and-test.sh  # [unix]
+        -   popd  # [unix]
+        -   pushd cmake
+        -     cmake -GNinja -DPY_VER={{ version }}
+              # --trace --debug-output --debug-trycompile .
+        -   popd
+        - popd
+        - test ! -f default.profraw   # [osx]
+  - name: libpython-static
```
scopatz, line 213: Shouldn't this just be `python-static` to follow the convention?

xhochy: That would be a bit confusing in this case as the main Python is also statically linking libpython. As it's only the lib what is included in the -static package, I would rather reflect that also in the name.

scopatz: It still seems a bit odd to me.  I am not blocking based on the name though.

xhochy: Any other opinions? I would happily sacrifice the name for a merge but I still think `libpython-static` is more intuitive regarding the contents of the new output.

recipe/build_static.sh
```diff
@@ -0,0 +1,25 @@
...
+if [[ ${DEBUG_PY} == yes ]]; then
+  DBG=d
+else
+  DBG=
+fi
+VER=${PKG_VERSION%.*}
+VERABI=${VER}${DBG}
+
+
+cp -pf ${_buildd_static}/libpython${VERABI}.a ${PREFIX}/lib/libpython3.8.a
```
mariusvniekerk, line 15: The 3.8 here probably shouldn't be explicit

xhochy: Fixed and CI passed successfully.

recipe/meta.yaml
```diff
@@ -116,83 +116,119 @@ build:
...
+        - posix  # [win]
+      host:
+        - {{ pin_subpackage('python', exact=True) }}
+      run:
+        - {{ pin_subpackage('python', exact=True) }}
+    test:
+      commands:
+        - test -f ${PREFIX}/lib/libpython3.8.a  # [unix]
+        - test -f ${PREFIX}/lib/libpython3.8.nolto.a  # [unix]
+        - test -f ${PREFIX}/lib/python3.8/config-3.8-darwin/libpython3.8.a  # [osx]
```
scopatz, line 231: explicit 3.8 mentions here

recipe/meta.yaml
scopatz: I am also happy to be outvoted here too.

xhochy:
@conda-forge/python ping again here, merging this would be a massive improvement as it should reduce the on-disk size of a basic Python installation by 25%.

Issue #270, opened 2019-08-25
Document and maintain Python pre-releases


willingc:
Hi team :wave:

I suspect that some maintainers would appreciate an easily installable version of Python pre-releases to demonstrate how their project works with a prerelease's new feature.

As I understand the process of creating a pre-release so far:

- Prereleases are found in named branches on the python-feedstock repo
- The recipe's [`meta.yaml`](https://github.com/conda-forge/python-feedstock/blob/pre-3.8/recipe/meta.yaml) contains configuration for the release version number and a label to indicate it is a pre-release.

I'm not clear on:

- the series of patches in the `meta.yaml` file
- how to render/rerender a pre-release branch so the label applies (does it do it automagically with the existing commands?)

If folks see value in this, I would be happy to iterate the details here and then document it. Thanks!

mingwandroid:
Hey Carol. Apologies for the silence around here! I'll respond in full a bit later.

willingc:
No worries @mingwandroid. Thanks for looking at it.

mingwandroid:
I actually wrote a big reply, with some steps to rebase patches but upon trying them they're not working at present so I well, deleted it all!

I will try to get back to you and hopefully we can get 3.8 previews ready before 3.8 is released (ten odd days, right? - and if not let's make a promise to do better for 3.9).

jjhelmus:
conda-forge/conda-forge.github.io#861 is a request for documentation on setting up pre/dev releases in conda-forge in general which would be beneficial here.

megies:
3.8.0rc1 is out, would be fantastic to get it built on conda-forge so projects can add it to their CI matrices

ocefpaf:
Python 3.9 is in alpha at the moment and we need to decide on a policy for our dev releases.

Due to the huge number of patches and the difficult to build python I don't think it is feasible for the maintainers of the feedstock to do all the dev releases (alpha, beta, and RCs). Maybe we should do just the RCs? After all conda-forge is a platform for users and one could argue that only RCs are really meant for user consumption. What do you think @willingc?

Pinging @CJ-Wright who brought this in our core meeting recently.

megies:
> Maybe we should do just the RCs? After all conda-forge is a platform for users and one could argue that only RCs are really meant for user consumption

Sounds very reasonable to me :+1:

PR #389, opened 2020-09-16
Python 3.9.0rc recipe

ocefpaf:
xref #270. @willingc You opened it but we never addressed this properly. The main issue is bandwidth to catch up with upstream, even if that will happen only on the RCs. I'm investigating the patches here and I hope to be able to work with upstream to see if we can get some of them merged (or dropped for something better).


recipe/meta.yaml
```diff
@@ -1,61 +1,279 @@
...
+      - patches/0006-Support-cross-compiling-byte-code.patch
+      - patches/0007-Win32-Fixes-for-Windows-GCC-interop-needed-by-RPy2-a.patch
+      - patches/0008-Darwin-Look-in-sysroot-usr-lib-include-if-sysroot-is.patch
+      - patches/0009-runtime_library_dir_option-Use-1st-word-of-CC-as-com.patch
+      - patches/0010-Win32-Do-not-download-externals.patch
+      - patches/0011-Add-support-for-_CONDA_PYTHON_SYSCONFIGDATA_NAME-if-.patch
+      - patches/0012-Fix-find_library-so-that-it-looks-in-sys.prefix-lib-.patch
+      - patches/0013-Disable-new-dtags-in-unixccompiler.py.patch
+      - patches/0014-Fix-cross-compilation-on-Debian-based-distros.patch
+      # - patches/0015-Disable-registry-lookup-unless-CONDA_PY_ALLOW_REG_PA.patch
```
ocefpaf, line 37: @mingwandroid this is inside a `if (!Py_IgnoreEnvironmentFlag)` now. Does that solve our problem or do we still need to refresh this patch.

willingc:
@ocefpaf No worries; we're all busy. Thanks for looking into it. If not this release, there will always be a new one :wink:

ocefpaf:
@mingwandroid all the patches are refreshed. It looks like we can drop `0021-roll-back-2e33ecd-for-py_compile.main.patch` and I'm not 100% sure I got `=0015-Disable-registry-lookup-unless-CONDA_PY_ALLOW_REG_PA.patch` correctly, but the rest seems fine.

Do you have an idea of which ones we should prioritize and try to push upstream?

I'll investigate the Windows failure later, not sure what is going on. Linux and macOS are building correctly, the problem seems to be the lack of `python_abi` metapackage for Python 3.9. (Thanks @isuruf for pointing that out to me.)

recipe/meta.yaml
mingwandroid: I highly recommend keeping this. Without it, installers will fail, for example. We don't set `Py_IgnoreEnvironmentFlag` and until we do (if we want to) and get this enforced in all the places we need to.

mingwandroid:
Hey @ocefpaf,

My thoughts on upstreaming/dropping, patch-by-patch..

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0000-Fix-off-by-one-error-in-_winapi_WaitForMultipleObjec.patch
.. is in review upstream already. There was some pings about my PR yesterday. We can drop this once merged. Without it, you cannot install Miniconda or Anaconda (or do various other things) on a 64+ thread Windows machine. John Carmack raised this bug with us on twitter!

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0001-Add-Anaconda-Distribution-version-logic.patch
Not appropriate for upstream unless someone fancies tackling a general way to hook this bannering stuff. Anyway, conda-forge do not apply this one AFAIR.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0002-Darwin-Add-arch-x86_64-to-UnicCCompiler-ld_args.patch
Probably needs fixing for osx-aarch64 (ping @isuruf) maybe it can be removed? Worth checking.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0003-Win32-Change-FD_SETSIZE-from-512-to-2048.patch
Should be pushed upstream so that upstream Python is similarly capable.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0004-Win32-distutils-Add-support-to-cygwinccompiler-for-V.patch
Good simple bug fix, may want to add vs2019 stuff to it and push it upstream.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0005-Do-not-pass-g-to-GCC-when-not-Py_DEBUG.patch
Not sure about this one. We may want to check the size of extension modules if we removed it and if there is a difference then think about an upstreaming strategy for it.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0006-Support-cross-compiling-byte-code.patch
This has been important when needing to bootstrap our Python on a different linux arch. I don't know if things are better upstream in this regard but since this patch rebases cleanly my firm suspicion is no, it hasn't. I expect the chances of upstream wanting to touch this sort of code to be low.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0007-Win32-Fixes-for-Windows-GCC-interop-needed-by-RPy2-a.patch
Good, simple fix, should be upstreamed IMHO.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0008-Darwin-Look-in-sysroot-usr-lib-include-if-sysroot-is.patch
We could try to build without this, if it works (for building a load of extensions without problem) then consider dropping it, otherwise we should think about upstreaming it. 

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0009-runtime_library_dir_option-Use-1st-word-of-CC-as-com.patch
This patch allows setting e.g. "CC=gcc -g0", or "CC=ccache gcc", we should perhaps try to upstream it and get some opinions on it. I wonder if for correctness, we should be splitting by IFS though instead of ' '.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0010-Win32-Do-not-download-externals.patch
Not appropriate for upstreaming, this is to speed up our builds by not downloading things we don't need.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0011-Add-support-for-_CONDA_PYTHON_SYSCONFIGDATA_NAME-if-.patch
Not appropriate for upstreaming unless someone wants to have a go at supporting this sort of thing upstream in a way that they're likely to approve of. I doubt they would want this though. It's just a necessary evil for us, IMHO.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0012-Fix-find_library-so-that-it-looks-in-sys.prefix-lib-.patch
This patch makes Python better respect HFS (executable_path/../lib) when looking for dylibs on macOS. We could propose it upstream but it is quite conda-ecosystem specific IMHO.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0013-Disable-new-dtags-in-unixccompiler.py.patch
Here, conda-build is opinionated (as am I) about the new-dtags feature of glibc/the GNU linker. This prevents DSO interposition on Linux which many consider a security risk. The comment gives more detail. I don't know how upstream would feel about it, probably some would like it and some would dislike it!

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0014-Fix-cross-compilation-on-Debian-based-distros.patch
This should be pushed upstream. Debian-ism paths shouldn't be looked in when cross compiling.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0015-Disable-registry-lookup-unless-CONDA_PY_ALLOW_REG_PA.patch
This is important and not upstreamable IMHO.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0016-Unvendor-openssl.patch
https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0017-Unvendor-sqlite3.patch
Conda-ism, this implements our preference to use our own shared libraries. Not appropriate for upstream.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0018-venv-Revert-a-change-from-https-github.com-python-cp.patch
Conda-isms. If we don't want to do this then we'd need to align our layout on Windows better with upstream. I expect that would break a lot of things.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0019-Win-Add-back-the-DLLSuffix-used-to-find-the-openssl-.patch
@mariusvniekerk? Looks like something upstream might consider. But we'd likely need to also change the OpenSLL vcxproject files (that we do not use but upstream does, I think!)

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0020-Use-ranlib-from-env-if-env-variable-is-set.patch
We should try to get this upstreamed. I don't think it's very controversial (though changing the build system of Python in any way is somewhat!)

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0021-roll-back-2e33ecd-for-py_compile.main.patch
@duncanmmacleod? Seems like a bug upstream to me and we could try to upstream this fix.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0022-Add-CondaEcosystemModifyDllSearchPath.patch
I would rather not have had to write any of this but changes to upstream made it necessary. If you don't use this you will have problems, or if you don't your friends on Windows will. For my perspective 0% chance of being upstreamed.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0023-Add-d1trimfile-SRC_DIR-to-make-pdbs-more-relocatable.patch
Upstream would likely consider this.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0024-Doing-d1trimfile.patch
.. and maybe (along with) this.

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0025-egg-debugging-with-Windows-debug-builds.patch
There's a bug in egg stuff on Windows, but no one ever uses it. This was something I added when investigating that but I've completely lost the thread of that work now! Drop?

https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0026-Revert-part-of-https-bugs.python.org-issue33895-http.patch
@isuruf and I are in alignment that we'll neuter this patch but make it something you can use via an env var to aid debugging.

So we need a plan of attack. I'm more than happy to ask for time to be set aside for me to handle pushing these, starting with the easy ones first and seeing how far I go, but anyone is (always) welcome to propose any of my patches and ping me in the comments if they wish.

mingwandroid:
I used AnacondaRecipes URLs in my last comment, sorry about that, but it's to make it easier for me when/if I tackle this.

duncanmmacleod:
> https://github.com/AnacondaRecipes/python-feedstock/blob/master/recipe/patches/0021-roll-back-2e33ecd-for-py_compile.main.patch
> @duncanmmacleod? Seems like a bug upstream to me and we could try to upstream this fix.

@mingwandroid, this is already fixed upstream, see [bug report](https://bugs.python.org/issue40456) and fix https://github.com/python/cpython/pull/17134. Actually, that specific patch isn't in 3.9.0rc1, but I've checked the contents of that tarball and the bug that my patch fixed isn't there, so this one can be dropped either way.

ocefpaf:
@mingwandroid thnaks for the compilation of the patch status. @duncanmmacleod indeed the 0021-roll-back-2e33ecd-for-py_compile.main.patch [seems to be in](https://github.com/conda-forge/python-feedstock/pull/389/files#diff-e178b687b10a71a3348107ae3154e44cR44), I commented it out and I'm planning of removing it next iteration.

I'm copying https://github.com/conda-forge/python-feedstock/pull/389#issuecomment-695186852 in a doc so we can tackle one at a time. We already have @willingc :eyes: here ;-p

mingwandroid:
@zooba may want to take a look too?

isuruf:
The test error is because of a `cython` dependency in tests. not sure why it was added. Removing `cython` should unblock the tests.

recipe/meta.yaml
```diff
@@ -167,7 +167,6 @@ outputs:
         - ripgrep
         - cmake
         - ninja
-        - cython
```
ocefpaf: Looks like that was not it, or not just it.

What should we do about https://github.com/conda-forge/python-feedstock/pull/389/files#diff-e178b687b10a71a3348107ae3154e44cR124?

recipe/meta.yaml
```diff
@@ -1,61 +1,281 @@
...
+      run:
+        - ld_impl_{{ target_platform }} >=2.34  # [linux]
+      run_constrained:
+        - python_abi 3.8.* *_cp38
+    test:
+      requires:
+        - ripgrep
+        - cmake
+        - ninja
+        - setuptools
```
isuruf, line 170: @ocefpaf, need to remove this too

recipe/meta.yaml
```diff
@@ -1,61 +1,281 @@
...
+        - if not exist %PREFIX%\\Scripts\\pydoc.exe exit 1  # [win]
+        - pushd tests
+        -   pushd distutils
+        -     python setup.py install -v -v
+        -     python -c "import foobar"
+        -   popd
+        -   pushd distutils.cext
+        -     python setup.py install -v -v
+        -     python -c "import greet; greet.greet('Python user')" | rg "Hello Python"
+        -   popd
```
isuruf, line 221: These 4 lines too

mingwandroid:
Rather than removing all these tests, can you elide them when the patch number is 0? I think that way, we can do a build 1 once the necessary deps are present and get the benefits of the tests back.

ocefpaf:
@isuruf and @mingwandroid we should probably document those circular dependencies and how we work around them.

There is also https://www.python.org/dev/peps/pep-0615/#search-path-configuration. Jonathan brought that to my attention (Thanks!). I guess we can just add it as a run dependency on Windows, what do you think? (I'm still reading the pep and the recommendations but looks like we don't need to set anything for Python to find the system ones on Linux and macOS.)

isuruf:
IMO, those tests should be moved to the respective feedstocks and they can be added here as `test_downstreams`.

mingwandroid:
> IMO, those tests should be moved to the respective feedstocks and they can be added here as `test_downstreams`.

That's fine but there's no need to remove them in the meantime. Every one of these tests were added because they caught serious regressions!

mingwandroid:
They should be elided via jinja2 for patch version 0 instead. Your idea to put them elsewhere is ok but I'd personally prefer they remained in the package they are testing.

recipe/build_base.bat
```diff
@@ -133,7 +135,7 @@ if errorlevel 1 exit 1
 
 :: Populate the libs directory
 mkdir %PREFIX%\libs
-copy /Y %SRC_DIR%\PCbuild\%BUILD_PATH%\python38%_D%.lib %PREFIX%\libs\
+copy /Y %SRC_DIR%\PCbuild\%BUILD_PATH%\python%PY_VER%%_D%.lib %PREFIX%\libs\
```
ocefpaf, line 136: :tada:

ocefpaf:
> IMO, those tests should be moved to the respective feedstocks and they can be added here as `test_downstreams`.

Is `test_downstreams` a thing in conda-build? (Sorry for the dumb question, I'm very outdated in its development.)

mingwandroid:
We could add an option to ignore tests where the dependencies are missing easily enough I guess?

`test_downstreams` could work, I've not used it enough personally to really vouch for it though. That'd imply moving these tests into other packages (1 would get added to cmake, for example, and a few to setuptools). I would be happy with something/anything to make it so the recipes do not need editing to remove the tests when we change minor versions. I'd also like conda-build to record somewhere in the `info` dir which tests got skipped due to dependency issues so we can feel happy that a package has been well tested (or know otherwise).

mingwandroid:
An example of `test/downstreams` (ok, removal thereof): https://github.com/conda-forge/perl-feedstock/commit/0aa2b9c9db5447e3ec1d45f12ebaa85c69ca068e

recipe/meta.yaml
```diff
@@ -1,61 +1,283 @@
...
+        - python3-config --help  # [not win]
+        - set "PIP_NO_BUILD_ISOLATION=False"  # [win]
+        - set "PIP_NO_DEPENDENCIES=True"  # [win]
+        - set "PIP_IGNORE_INSTALLED=True"  # [win]
+        - set "PIP_NO_INDEX=True"  # [win]
+        - set "PIP_CACHE_DIR=%CONDA_PREFIX%/pip_cache"  # [win]
+        - set "TEMP=%CONDA_PREFIX%/tmp"  # [win]
+        - mkdir "%TEMP%"  # [win]
+        - python -Im ensurepip --upgrade --default-pip  # [win]
+        - python -c "from zoneinfo import ZoneInfo; from datetime import datetime; datetime(2020, 10, 31, 12, tzinfo=ZoneInfo('America/Los_Angeles')); print(dt.tzname())"
```
ocefpaf, line 197: I expected this to fail only on Windows but it seems that our macOS and Linux images do not have the zoneinfo data. Question? Should we add `tzdata` to all platforms or should we add it only on Windows and do not perform this tests on the other platforms?

Most of the time users will have that install by their systems already and according to https://www.python.org/dev/peps/pep-0615 Python will be smart enough to find it and prioritize it. What do you think @isuruf and @mingwandroid?

recipe/meta.yaml
```diff
@@ -1,61 +1,284 @@
...
+    build:
+      activate_in_script: True
+      # Windows has issues updating python if conda is using files itself.
+      # Copy rather than link.
+      no_link:
+        - DLLs/_ctypes.pyd  # [win]
+      # We delete the shared libraries.
+      ignore_run_exports:   # [unix]
+        - bzip2             # [unix]
+      # Disabled until new conda-build release
```
ocefpaf, line 101: @mingwandroid and @isuruf do you know what is the status of this? Can we just remove the commented out lines below?

recipe/meta.yaml
```diff
@@ -1,61 +1,284 @@
...
+      skip_compile_pyc:
+        - '*.py'            # [osx and arm64]
+      # Would like to be able to append to the calculated build string here, ping @msarahan, is this possible?
+      string: h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}{{ linkage_nature }}{{ debug }}_cpython
+      run_exports:
+        noarch:
+          - python
+        weak:
+          - python_abi 3.8.* *_cp38
+    # Not until: https://github.com/mingwandroid/conda-build/commit/c68a7d100866df7a3e9c0e3177fc7ef0ff76def9
```
ocefpaf, line 123: @mingwandroid is this upstream in conda-build? Can we uncomment the lines below?

mingwandroid: Yes it is. Setting this variable should work ok. I need to add a check to bld.bat for defaults but I'll do that when I recsync.

ocefpaf: Thanks!

ocefpaf:
> An example of `test/downstreams` (ok, removal thereof): [conda-forge/perl-feedstock@0aa2b9c](https://github.com/conda-forge/perl-feedstock/commit/0aa2b9c9db5447e3ec1d45f12ebaa85c69ca068e)

If I understood if correctly, here in this recipe, we would need https://github.com/conda-forge/python-feedstock/pull/389/commits/9b312a72c65e894051ed26a819aeef1f1b025c58, right? But the cython and setuptools tests here must be moved to cython and setuptools itself. Is that correct? conda-build will run those tests here of possible?

mingwandroid:
That's the hope yeah. If the downstream packages have good enough tests then we don't strictly need to move these tests to those packages.

recipe/meta.yaml
```diff
@@ -1,61 +1,277 @@
...
+        - ncurses  # [linux]
+        - libffi  # [not win]
+        - ld_impl_{{ target_platform }} >=2.34  # [linux]
+      run:
+        - ld_impl_{{ target_platform }} >=2.34  # [linux]
+        - tzdata
+      run_constrained:
+        - python_abi 3.8.* *_cp38
+    test:
+      downstream:
```
isuruf, line 164:
```suggestion
      downstreams:
```

ocefpaf: Thanks!

ocefpaf:
Not sure what is going on here now. Could it be the new conda-build? Pinging @mingwandroid and @isuruf.

mingwandroid:
I'm looking at the patch failures. Can I ask that you anyone who adds patches here numbers them in the pre-existing scheme?

```
      - patches/gh21114.patch
      - patches/gh21241.patch
      - patches/gh21228.patch
      - patches/gh21564.patch  # [osx and arm64]
```

To fix patches, I use git am -3 <dir>/*.patch and if they are named like this then the order of that is not controllable or reproducible.

Also, please if we can, avoid eliding patches depending on the target platform. This is not a sustainable or upstreamable way to handle patches.

ocefpaf:
I agree that those patches require some work. Let me see what I can do here.

mingwandroid:
Also, any idea why 0015- ... has lost its git header/formatting compared to mine?

```
From 0a69599f414efc5811b5143d37c639fa7ed6123b Mon Sep 17 00:00:00 2001
From: Ray Donnelly <mingw.android@gmail.com>
Date: Sat, 27 Oct 2018 18:48:30 +0100
Subject: [PATCH 15/26] Disable registry lookup unless CONDA_PY_ALLOW_REG_PATHS
 is not 0

---
 PC/getpathp.c | 14 ++++++++++++--
 1 file changed, 12 insertions(+), 2 deletions(-)

diff --git a/PC/getpathp.c b/PC/getpathp.c
index 8969fb554a..4a9e47ae45 100644
--- a/PC/getpathp.c
+++ b/PC/getpathp.
```

ocefpaf:
> Also, any idea why 0015- ... has lost its git header/formatting compared to mine?

Mostly b/c I use another tool. I need to learn how you manage them to reduce the amount of changes.
If you have some time I'd love if you could refresh these patches in the way you maintain your recipes, if not, that is the best I can do at the moment.

mingwandroid:
I can for sure go through how I rebase patches. It's just using git though. I am starting to add CONDA_BUILD_FROM_SRC_CONTROL env var checking to my recipes (the tricky ones with patches) which meta.yaml will use to fetch the sources from git.

If conda-build can fetch all sources from git and all patches are git patches then it will use git to apply those patches to $SRC_DIR. If it does this and there's a failure to apply the patches, you are left with a useful working tree with the patches that did apply ok, applied and the current failing one left in a state that allows manual fixing.

As soon as any of the git headers get dropped this (very good) workflow cannot, err, work!

I spent a while trying to update the patches my way before I noticed that too many of them have dropped this info. It's also important from a credit/blame perspective that we don't drop headers.

If people don't want to use git format-patch then you can still manually re-patch the patches so that this info is retained. That's a bit fiddly though.

mingwandroid:
Also, git format-patch uses the sha info to help it do a good job rebasing patches. This is very handy (when it works!)

ocefpaf:
The weight of maintaining that huge amount of patches in a workflow that satisfies all those criteria is huge. I cannot ask anyone to do that. While giving credit for the patch is great we are actually doing it twice when using github to maintain the recipe and the patches. I can try to use your workflow but it will take a while for me to learn it and apply it here.

mingwandroid:
It really, really is *not*. It just requires knowing how to use a few git commands!

I updated the AnacondaRecipes version of all these patches. It took 5 minutes.

I wasted 2 hours trying to apply the patches that have been mangled here :-(

ocefpaf:
> I updated the AnacondaRecipes version of all these patches. It took 5 minutes.
> 
> I wasted 2 hours trying to apply the patches that have been mangled here :-(

That is the thing about the tool we know, right?

I wasted two days updating these patches in the way they were made. While I usually also just do that in <2min with my tools. TL;DR I can try to learn your way but I cannot do it today.

mingwandroid:
IMHO we should prohibit changes that drop credit and/or comments from patches.

mingwandroid:
Here are my rebased patches. Builds succeed on macOS and Linux but the solver fails to solve when trying to run the embedded tests.

https://github.com/AnacondaRecipes/python-feedstock/tree/python-3.9rc2/recipe/patches

ocefpaf:
> IMHO we should prohibit changes that drop credit and/or comments from patches (header blocks in general).

I agree. But like I said the credit is not gone from the history, just from a header that exists in a tool of choice. Anyways, I don't really care as long as we have people that know the tools and are willing to use them. Otherwise I will take whatever new maintainers what do to that makes their life easier.

mingwandroid:
Do you want me to push a PR that updates these shared patches? Can I push that here?

mingwandroid:
Oh I updated the against rc2!

mingwandroid:
https://github.com/ocefpaf/python-feedstock/pull/3

mingwandroid:
You can now set `CONDA_BUILD_FROM_SOURCE_CONTROL` env. var, and `conda-build` will use `git`.

I also disabled `--pgo-extended`. I reckon this is badly named as it gives the impression you'll get better PGO results. That isn't the case. The tests run by `--pgo` are considered useful for PGO and those excluded considered noise-y or useless for PGO (it's just the full regression testsuite!)

mingwandroid:
Looks like I broke Windows. I'll look at it tomorrow. Good night.

mingwandroid:
https://github.com/ocefpaf/python-feedstock/pull/4/

mingwandroid:
https://github.com/ocefpaf/python-feedstock/pull/5

ocefpaf:
> [ocefpaf#5](https://github.com/ocefpaf/python-feedstock/pull/5)

You can send commits direct to this PR if you want. That will cut the middle-merge-man :smile:

recipe/meta.yaml
mingwandroid: Actually I am not sure. I mean, this is our own var that some may have used.. if no one is using it then yeah, let's drop it.

mingwandroid:
This builds now. What's the plan with tzdata though?

mingwandroid:
Does anyone know why the linter is complaining? Each output has build/number: 0 now.

mingwandroid:
This doesn't address the tzdata thing. We can do that in a follow up PR?

ocefpaf:
@mingwandroid are you OK pointing this to the rc label, switching to the rc2, and merging it there first? That way we can wait for the tzdata stuff and prepare this for master in another PR.

ocefpaf:
> This builds now. What's the plan with tzdata though?

#392

@mingwandroid awesome work! We will merge this in the rc label for now, without the tzdata, and I'll port this recipe to `master` as soon as we have https://github.com/conda-forge/tzdata-feedstock/pull/1

mingwandroid:
Nice. Thanks

PR #390, opened 2020-09-17
Remove unnecessary patches

isuruf:
Checklist


isuruf:
@mingwandroid now that https://github.com/conda-forge/lief-feedstock/pull/12 is merged this patch is unnecessary

mingwandroid:
For patch 0026, I believe that this issue will crop up in other extension modules built with pybind11 on Windows and rather than removing the patch I'd prefer that we reverse the sense of it so that it does nothing by default, but can be activated by setting `CONDA_DLL_SEARCH_MODIFICATION_DO_NOT_DROP_GIL=yes` instead. So it would turn into a debugging tool for us only (and an emergency workaround for people). Your thoughts are very welcome @isuruf.

isuruf:
Sounds good. I'll send a PR if I find time

mingwandroid:
I don't mind if you proceed, then next time I sync back (hopefully soon), I can do this work, but more than happy to share that load with you too, I guess we'll see who gets there first.

PR #393, opened 2020-10-01
Name feedstock python again

### LINKED ISSUE #382: What is the difference between python-split-feedstock and python-feedstock?
What is the difference between python-split-feedstock and python-feedstock? The README mentions python-split in a few places.

xhochy:
Fixes #382

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-About python-split
-==================
+About python
+============
 
 Home: https://www.python.org/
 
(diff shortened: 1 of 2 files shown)


xhochy:
Weird that patches don't apply anymore here :(

ocefpaf:
> Weird that patches don't apply anymore here :(

Yeah, it is the new conda-build. We need to do them in a certain way now or they won't work.

xhochy:
> > Weird that patches don't apply anymore here :(
> 
> Yeah, it is the new conda-build. We need to do them in a certain way now or they won't work.

Is there any documentation on what this means? I don't see anything in `conda-build`'s changelog.

ocefpaf:
> Is there any documentation on what this means? I don't see anything in `conda-build`'s changelog.

[This is all I got](https://github.com/conda-forge/python-feedstock/pull/389#issuecomment-702090982).

mbargull:
@mingwandroid, looks like one of the patches has multiple commits on the same file. And `patch --dry-run` doesn't seem smart enough to handle that :thumbsdown:. Would it be possible to just use `git apply --check`? `git apply` works outside of Git repos and should be more capable that `patch` -- and `conda-build` uses `git` anyway for some of its functionality\*.

(\* The `conda-build` packages still don't declare a dependency on Git, but IMO we should change that.)

mingwandroid:
The best pattern to follow is to have to level requirements/build with any source code fetching tools on it, then use multiple outputs with their own requirements/build which do not need the source fetching tools.

You can see this done on the python 3.9 recipe now.

I hi have nearly finished the 3rd go at reworking how patches are applied. I'll hopefully release it tonight or tomorrow.

Cheers.

PR #395, opened 2020-10-06
Py39

### LINKED ISSUE #392: How should we handle tzdata/zoneinfo
For now I commented out the `tzdata` package and the new `datetime` functionality won't work in our RC release. See https://github.com/conda-forge/python-feedstock/pull/389/commits/dd8de1999842e015dab76b06020e61d465f45535

We should figure out a way to provide a python `tzdata` package for Windows and Linux/macOS for testing avoiding the circular dependencies or complex builds.

The alternatives are:

- Python a special case py39 without it and then build tzdata
- Create a non-python tzdata package that would not require python at install time
- Don't test it and let users rely on whatever the system provides (this will leave Windows users hanging though!)

xref.: #389

ocefpaf:
Ping @isuruf 

Needs https://github.com/conda-forge/tzdata-feedstock/pull/1
Closes #392


recipe/meta.yaml
```diff
@@ -230,17 +293,30 @@ outputs:
...
+      number: 0
       activate_in_script: True
-      # Would like to be able to append to the calculated build string here, ping @msarahan, is this possible?
+      ignore_run_exports:
+        - python_abi
+        - clangxx_osx-64  # [osx64]
+        - clangxx_osx-aarch64  # [osxaarch64]
+        - gxx_linux-64  # [linux64]
+        - gxx_linux-aarch64  # [linuxaarch64]
+        - gxx_linux-ppc64le  # [linuxppc64le]
```
isuruf, line 307: These don't do anything.
```suggestion
```

recipe/patches/0001-Darwin-Add-arch-x86_64-to-UnicCCompiler-ld_args.patch
```diff
@@ -1,25 +1,24 @@
...
 --- a/Lib/distutils/unixccompiler.py
 +++ b/Lib/distutils/unixccompiler.py
-@@ -200,6 +200,8 @@ class UnixCCompiler(CCompiler):
+@@ -200,6 +200,7 @@ class UnixCCompiler(CCompiler):
  
                  if sys.platform == 'darwin':
                      linker = _osx_support.compiler_fixup(linker, ld_args)
-+                    from distutils.util import get_host_platform
-+                    ld_args = ['-arch', get_host_platform().split("-")[-1]] + ld_args
++                    ld_args = ['-arch', 'x86_64'] + ld_args
```
isuruf, line 18: Can you revert this change?

ocefpaf: Sure. Working on it...
I just wonder why we needed that in the first place, Maybe @mingwandroid can chime-in.

recipe/patches/gh21228.patch
```diff
@@ -1,47 +0,0 @@
-From 79d170ebfa523f4a4f104d8041ad880215c5d5fa Mon Sep 17 00:00:00 2001
```
isuruf, line 1: Need to restore this patch

ocefpaf: Done. Not sure what is going on with the conflicts :-(

recipe/patches/0001-Darwin-Add-arch-x86_64-to-UnicCCompiler-ld_args.patch
mingwandroid: To do with fat objects back when the tools supported (and defaulted) to i686 AFAIR, probably fine to jettison at this point.

recipe/sysconfigdata/default/_sysconfigdata_linux.py
```diff
@@ -651,36 +710,35 @@
...
- 'TESTRUNNER': 'LD_LIBRARY_PATH=/opt/anaconda1anaconda2anaconda3/../work/Python-@PYVERFULL@ '
-               './python ./Tools/scripts/run_tests.py',
+ 'TESTRUNNER': './python '
+               './Tools/scripts/run_tests.py',
  'TESTTIMEOUT': 1200,
- 'THREADOBJ': 'Python/thread.o',
  'TIMEMODULE_LIB': 'rt',
  'TIME_WITH_SYS_TIME': 1,
  'TM_IN_SYS_TIME': 0,
+ 'TZPATH': '/usr/share/zoneinfo:/usr/lib/zoneinfo:/usr/share/lib/zoneinfo:/etc/zoneinfo:/opt/anaconda1anaconda2anaconda3/share/zoneinfo',
```
ocefpaf, line 732: Editing this is quite cumbersome and error prone. I wonder if there is a better way or if we are stuck with this :-/

mingwandroid:
Yes, I developed a replacements thing recently just to get away from having to bash script so much brittle stuff. Here's an example of it: https://github.com/AnacondaRecipes/qt-feedstock/blob/master/recipe/conda_build_config.yaml#L9-L28

recipe/meta.yaml
```diff
@@ -221,7 +221,7 @@ outputs:
         - ld_impl_{{ target_platform }}  # [linux]
       run:
         - ld_impl_{{ target_platform }}  # [linux]
-        - tzdata
+        - python-tzdata
```
ocefpaf, line 224: I thought that the idea was to avoid the python package and force Python to find the `tzdata` noarch generic, no?

mingwandroid:
Ah maybe I made a mistake.

mingwandroid:
I think we want to use python-tzdata here, otherwise we get:

```
Python 3.9.0 | packaged by conda-forge | (default, Oct  6 2020, 17:37:43) 
[Clang 10.0.1 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from zoneinfo import ZoneInfo
>>> from datetime import datetime, timedelta
>>> dt = datetime(2020, 10, 31, 12, tzinfo=ZoneInfo("America/Los_Angeles"))
Traceback (most recent call last):
  File "/opt/conda/envs/py390.cf/lib/python3.9/zoneinfo/_common.py", line 12, in load_tzdata
    return importlib.resources.open_binary(package_name, resource_name)
  File "/opt/conda/envs/py390.cf/lib/python3.9/importlib/resources.py", line 88, in open_binary
    package = _get_package(package)
  File "/opt/conda/envs/py390.cf/lib/python3.9/importlib/resources.py", line 49, in _get_package
    module = _resolve(package)
  File "/opt/conda/envs/py390.cf/lib/python3.9/importlib/resources.py", line 40, in _resolve
    return import_module(name)
  File "/opt/conda/envs/py390.cf/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 972, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 972, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 984, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'tzdata'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/conda/envs/py390.cf/lib/python3.9/zoneinfo/_common.py", line 24, in load_tzdata
    raise ZoneInfoNotFoundError(f"No time zone found with key {key}")
zoneinfo._common.ZoneInfoNotFoundError: 'No time zone found with key America/Los_Angeles
```

mingwandroid:
@mbargull?

mingwandroid:
osx-arm gives:
> 2020-10-07T12:33:30.4130970Z /Users/runner/miniforge3/conda-bld/python-split_1602073181968/work/build_base.sh: line 401: 33344 Segmentation fault: 11  ${STRIP} -S lib/libpython${VERABI}.a

Shall we skip stripping there?

ocefpaf:
> Ah maybe I made a mistake.

Using python-tzdata will definitely fix it but I guess the idea is to avoid that and use `--with-tzpath`. However, that was not working and @isuruf made it work with [this change](https://github.com/conda-forge/python-feedstock/pull/395/commits/cd61f4eb340762d2e65600a3056a1ac559154f57#diff-7b6578aaf24f33556255377d2a5d56bcR903) but only on Linux.

recipe/patches/0032-Fix-TZPATH-on-windows.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/Lib/sysconfig.py b/Lib/sysconfig.py
+index bf04ac541e..11e81d4f44 100644
+--- a/Lib/sysconfig.py
++++ b/Lib/sysconfig.py
+@@ -546,7 +546,7 @@ def get_config_vars(*args):
+ 
+         if os.name == 'nt':
+             _init_non_posix(_CONFIG_VARS)
+-            _CONFIG_VARS['TZPATH'] = ''
++            _CONFIG_VARS['TZPATH'] = os.path.join(_PREFIX, "share", "zoneinfo")
```
ocefpaf, line 19: :tada:

ocefpaf:
@mingwandroid any final thoughts/consideration? Are we good to merge?

isuruf:
@mingwandroid, I removed the hand edited sysconfigdata files as they were outdated and added code to generate the sysconfigdata for the system gcc and the conda compiler. Let me know if you are okay with that.

mingwandroid:
I say go for it (then take cover)

mingwandroid:
Oh one quick thing:

case "$target_Platform" in

does the case of the P matter here?

isuruf:
yes, it does. fixed in f988bc854c33bdb1bea6b4e14c43ec79a3b80dab

ocefpaf:
> I say go for it (then take cover)

isuruf:
Thanks @mingwandroid and @ocefpaf

ocefpaf:
And you @isuruf! This one really took a village to get done!

ocefpaf:
Also thanks to @pganssle for guiding us with the tzdata package and @mbargull for creating it!

mingwandroid:
Great to see this land. From my perspective, it's the first time I've worked on an update to Python within conda-forge in the first instance. That's of course the best way, esp. for the important packages.

It was a huge effort for many of us and of course its completion brings a whole bunch of other stuff to build, but with conda-forge's excellent tooling (and GitHub) that's not too scary a prospect.

For now, I must get 3.9.0 on defaults ASAP (I added tzdata and python-tzdata already - do we need both still? I got a bit confused around that bit!). I'll base it on the recipe here, and rebase the anaconda advertising patch to happen last (this makes patch rebasing easier). On that note, we should also talk about adopting the anaconda advertising method for conda-forge too instead of the conda-forge branding stuff, it is more reliable and involves less file IO at interepreter launch time, pinging @jjhelmus on the reliability matter). CMake detects our tools now (thanks KitWare!) and will increasingly do so so having some more uniformity around this stuff would be good.

Keep on keeping on conda-forge.

ocefpaf:
> For now, I must get 3.9.0 on defaults ASAP (I added tzdata and python-tzdata already - do we need both still?

To build python we only need `tzdata`. We also offer `python-tzdata` b/c some packages may want to use it. They do share the same zoneinfo data now.

> I got a bit confused around that bit!). I'll base it on the recipe here, and rebase the anaconda advertising patch to happen last (this makes patch rebasing easier). On that note, we should also talk about adopting the anaconda advertising method for conda-forge too instead of the conda-forge branding stuff, it is more reliable and involves less file IO at interepreter launch time, pinging @jjhelmus on the reliability matter). CMake detects our tools now (thanks KitWare!) and will increasingly do so so having some more uniformity around this stuff would be good.

Sure. I don't really know what that entails but it sounds like something we could do.

mbargull:
> Great to see this land. From my perspective, it's the first time I've worked on a big update to Python within conda-forge in the first instance. That's of course the best way, esp. for the important packages.

:tada:

> (I added tzdata and python-tzdata already - do we need both still? I got a bit confused around that bit!).

Yep, like Felipe said, it makes sense to have both. `python-tzdata` isn't really important for *our* Python distribution, because with `tzdata` we should ship everything `zoneinfo` needs. But we'll probably see (few?) other packages that may add `python-tzdata` as a dependency in the future.

> Keep on keeping on conda-forge.

:metal:

----
> Sure. I don't really know what that entails but it sounds like something we could do.

Same here, agreed! Best thing would probably be to just open a PR whenever you're ready and discuss details there.

Superharz:
Hello,

I'm running on a 64-Bit Windows 10 v. 1909

I just tried to install python 3.9.0 in a new empty Conda Env (named new2).
However, I always get the following error:

grafik

mingwandroid:
You should use: `conda config --set add_pip_as_python_dependency false`

.. at least until pip is rebuilt.

(everyone should use this, always IMHO)

Superharz:
> 
> 
> You should use: `conda config --set add_pip_as_python_dependency false`
> 
> .. at least until pip is rebuilt.
> 
> (everyone should use this, always IMHO)

@mingwandroid Thank you. That solved my problem.

PR #396, opened 2020-10-07
V3.8.6

ocefpaf:
@mingwandroid I used your workflow to handle the patches here. Do you mind taking a look?


recipe/meta.yaml
```diff
@@ -268,6 +270,7 @@ about:
   dev_url: https://docs.python.org/devguide/
 
 extra:
+  feedstock-name: python
```
ocefpaf, line 273: @xhochy I incorporated https://github.com/conda-forge/python-feedstock/pull/393 here, is that OK?

mingwandroid:
I also have git-format-patch-le.py which is a little better than make-mixed in that it can emit patch files with the correct line endings when those vary between files. Still, @jjhelmus has good luck with `git config --global core.autocrlf false` for regenerating patches like this. I never had any luck and false is the default for that and I'm not sure it does anything on non-Windows.

recipe/meta.yaml
xhochy: Yes, sure!

ocefpaf:
> I also have git-format-patch-le.py which is a little better than make-mixed in that it can emit patch files with the correct line endings when those vary between files.

I have both scripts here. Thanks! I'm just experimenting for now. But good to know that `git-format-patch-le.py` is better.

> Still, @jjhelmus has good luck with `git config --global core.autocrlf false` for regenerating patches like this. I never had any luck and false is the default for that and I'm not sure it does anything on non-Windows.

Yep. I forgot to do that. But I'm adding it to the docs I'm writing.

jjhelmus:
LGTM

jjhelmus (approved):

jjhelmus:
Might be good change this PR to merge into the `3.8` branch so that #395 does not need to be re-based.

ocefpaf:
> Might be good change this PR to merge into the `3.8` branch so that #395 does not need to be re-based.

Indeed. Let me fix that.

ocefpaf:
Thanks @jjhelmus for the review! And thanks you both for teaching me something new! I now (kind of) know how to manage these patches.

PR #401, opened 2020-10-10
Fix the rebase of a patch

isuruf:
Checklist

### DIFF
--- a/recipe/patches/0024-Add-CONDA_DLL_SEARCH_MODIFICATION_KEEP_GIL-to-aid-de.patch
+++ b/recipe/patches/0024-Add-CONDA_DLL_SEARCH_MODIFICATION_KEEP_GIL-to-aid-de.patch
@@ -31,7 +31,7 @@ index 1b3e275614..071c2623b1 100644
  
          /* Don't display a message box when Python can't load a DLL */
          old_mode = SetErrorMode(SEM_FAILCRITICALERRORS);
-@@ -191,10 +196,19 @@ dl_funcptr _PyImport_FindSharedFuncptrWindows(const char *prefix,
+@@ -191,10 +196,15 @@ dl_funcptr _PyImport_FindSharedFuncptrWindows(const char *prefix,
          extern int CondaEcosystemModifyDllSearchPath(int, int);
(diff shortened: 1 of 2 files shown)


recipe/patches/0024-Add-CONDA_DLL_SEARCH_MODIFICATION_KEEP_GIL-to-aid-de.patch
```diff
@@ -48,10 +48,6 @@ index 1b3e275614..071c2623b1 100644
 +            hDLL = LoadLibraryExW(wpathname, NULL,
 +                                LOAD_WITH_ALTERED_SEARCH_PATH);
 +        }
-+
-+#if !USE_UNICODE_WCHAR_CACHE
-+        PyMem_Free(wpathname);
-+#endif /* USE_UNICODE_WCHAR_CACHE */
```
mbargull, line 54: Looks like this might've slipped in via applying parts of https://github.com/python/cpython/commit/4c8f09d7cef8c7aa07d5b5232b5b64f63819a743#diff-fa4616621faa1cbc0e83ef8930bb3202.
Which means in the future that "free" call is correct, but only if the pointer is not created via [`_PyUnicode_AsUnicode`](https://github.com/python/cpython/blob/v3.9.0/Include/cpython/unicodeobject.h#L576-L588) (<-- says "Return a read-only pointer ...").

(Just writing this down here for future (read 3.10 release next year) reference when the patch needs to be rebased.)

mbargull (approved):

recipe/patches/0024-Add-CONDA_DLL_SEARCH_MODIFICATION_KEEP_GIL-to-aid-de.patch
mariusvniekerk: And this is why I like nilling pointers so that it's easy (although slightly slower) to avoid this :)

isuruf: What's that?

PR #448, opened 2021-02-10
Re-sync with Anaconda Recipes WIP

mingwandroid:
Checklist


mingwandroid:
Please give your thoughts on this. I think it needs splitting, personally! Cheers all.

recipe/build_base.bat
```diff
@@ -2,8 +2,10 @@ setlocal EnableDelayedExpansion
 echo on
 
 :: brand Python with conda-forge startup message
-%SYS_PYTHON% %RECIPE_DIR%\brand_python.py
-if errorlevel 1 exit 1
+if "%CONDA_FORGE%"=="yes" (
+  %SYS_PYTHON% %RECIPE_DIR%\brand_python.py
+  if errorlevel 1 exit 1
+)
```
jakirkham, line 8: Can this script just be replaced in the Anaconda case? IOW is the condition needed?

recipe/build_base.bat
```diff
@@ -47,7 +49,13 @@ if "%DEBUG_C%"=="yes" (
...
 
 :: AP doesn't support PGO atm?
-set PGO=
+if "%CONDA_FORGE%"=="yes" (
+  set PGO=
+)
+:: PGO is not yet working for Python 3.9:
+:: (CopyPGCFiles target) ->
+::   C:\opt\conda\conda-bld\python-split-3.9.0_7\work\PCbuild\pyproject.props(165,5): error : PGO run did not succeed (no python39!*.pgc files) and there is no data to merge [C:\opt\conda\conda-bld\python-split-3.9.0_7\work\PCbuild\pythoncore.vcxproj]
+:: set PGO=
```
jakirkham, line 58: Maybe it makes sense to put this in a `conda_build_config.yaml` to make it easier for us both to configure?

recipe/build_base.sh
```diff
@@ -41,11 +39,13 @@ else
...
-fi
-if [[ ${target_platform} == linux-ppc64le ]]; then
-  _OPTIMIZED=no
+if [[ ${CONDA_FORGE} == yes ]]; then
+  if [[ ${target_platform} == linux-aarch64 ]]; then
+    _OPTIMIZED=no
+  fi
+  if [[ ${target_platform} == linux-ppc64le ]]; then
+    _OPTIMIZED=no
+  fi
```
jakirkham, line 48: Also seems like a good candidate for a `conda_build_config.yaml` value

jakirkham (commented):
Had some questions above about splitting out config values into `conda_build_config.yaml` to make these easier for us to set in different builds as needed. Also question about how branding is handled in the Anaconda case

recipe/build_base.bat
mingwandroid: Being baked into the C code, the anaconda method is likely faster and I believe our performance lead in startup tests is largely down to this. Still, we do check for a file existing to allow us to further customize the branding and conda-forge probably do not need/want that. So CF could retake this lead!

recipe/build_base.bat
mingwandroid: Locally to the recipe I guess?

recipe/build_base.sh
mingwandroid: Same question.

mingwandroid:
Thanks @jakirkham

recipe/build_base.bat
jakirkham: Yeah that's what I was thinking. Though open to other thoughts

mingwandroid: I am more than happy to change to this then.

mingwandroid: Upper case in CBC though or not? Shall I just go with OPTIMIZE and PGO?

recipe/build_base.bat
mingwandroid: Actually, the conda-forge method does not incur extra run-time overhead, I just checked. The one thing it does do though is to patch the source code inside the build scripts and I do prefer patching is done via patch files as it's more obvious what's been changed that way.

recipe/build_base.bat
jakirkham: Cool. No strong preference. Yeah that sounds reasonable. Thanks Ray! 😄

mingwandroid: Ah the one downside with putting it into CBC is that I cannot differentiate a default between CF and AD in the same file so sharing is more difficult. For example AD uses PGO on Windows but CF cannot at present. I have the same concern about OPTIMIZE where conda-forge have it disabled for ppc64le but I haven't seen the need to do that. Again, this is encoded into the recipe ATM so I can determine the appropriate value at runtime.

recipe/meta.yaml
```diff
@@ -280,19 +294,31 @@ outputs:
...
         - popd
         - test ! -f default.profraw   # [osx]
 
+  - name: python-regr-testsuite
+    requirements:
+      host:
+        - {{ pin_subpackage('python', exact=True) }}
+    script: install_regr_testsuite.sh  # [unix]
+
+{% if os.getenv("build_platform") != "linux-aarch64"  %}
```
isuruf, line 316: Why?

recipe/meta.yaml
```diff
@@ -280,19 +294,31 @@ outputs:
...
         -   popd
-        -   pushd prefix-replacement  # [unix]
-        -     bash build-and-test.sh  # [unix]
+        -   pushd embedding-interpreter  # [unix]
+        -     bash build-and-test.sh  # [unix and not (linux and aarch64)]
         -   popd  # [unix]
-        -   pushd cmake
-        -     cmake -GNinja -DPY_VER={{ version }}
-              # --trace --debug-output --debug-trycompile .
+        -   pushd cmake  # [not (linux and ppc64le)]
```
isuruf, line 300: Why?

recipe/build_static.sh
```diff
@@ -11,10 +11,21 @@ fi
...
+    ;;
+  *)
+    OLD_HOST=$HOST
+    ;;
+esac
 
 cp -pf ${_buildd_static}/libpython${VERABI}.a ${PREFIX}/lib/libpython${VERABI}.a
 if [[ ${HOST} =~ .*linux.* ]]; then
-  pushd ${PREFIX}/lib/python${VERABI}/config-${VERABI}-${HOST/-conda/}
+  pushd ${PREFIX}/lib/python${VERABI}/config-${VERABI}-${OLD_HOST}
```
isuruf, line 28:
```suggestion
  pushd ${PREFIX}/lib/python${VERABI}/config-${VERABI}-${OLD_HOST/-conda/}
```

recipe/build_base.sh
```diff
@@ -450,11 +496,9 @@ pushd "${PREFIX}"/lib/python${VER}
   # Append the conda-forge zoneinfo to the end
   sed -i.bak "s@zoneinfo'@zoneinfo:$PREFIX/share/tzinfo'@g" sysconfigfile
   # Remove osx sysroot as it depends on the build machine
-  sed -i.bak "s@-isysroot @@g" sysconfigfile
-  sed -i.bak "s@$CONDA_BUILD_SYSROOT @@g" sysconfigfile
+  # sed -i.bak "s@-isysroot ${CONDA_BUILD_SYSROOT}@@g" sysconfigfile
```
isuruf, line 499: Why?

recipe/meta.yaml
mingwandroid: I'm not exactly sure what you are asking here @isuruf.

This test was originally going to about prefix-replacement but it is not about that anymore so it's been renamed. Now it's a very comprehensive test to make sure we can embed both static and shared python interpreters in other programs.

recipe/meta.yaml
mingwandroid: Are you asking why we want to make a split package for the python regression testsuite here? Before we had split packages, Jonathan felt that the test-suite added too much to the core interpreter, so this is a step in the direction towards being able to better test our releases.

recipe/build_base.sh
mingwandroid: Because: https://github.com/conda-forge/python-feedstock/pull/448/files#diff-e0638ff31f64a715c0ed3faa90c7b9f61ef3ff28d4a622e89e31923e585394bbR387-R388

FWIW, the method did not work depending on the delineation / line wrapping that python was applying to sysconfig.py.

conda-build has its own way of doing replacements based on regexs now that I want to use instead of this stuff though and I will work later (if I can) on using that new method. Currently I used it in our Qt5 recipe and it works well there.

Can you be more detailed with your question though, I would like some more context if possible.

recipe/meta.yaml
mingwandroid: The cmake test is disabled for aarch64 atm because our cmake is too old there I think @katietz ?

isuruf: Yes, sorry. I meant to ask why it's skipped for `ppc64le`?

recipe/build_base.sh
isuruf: > FWIW, the method did not work depending on the delineation / line wrapping that python was applying to sysconfig.py.

Yes, that's why we had the two lines. I think you might have been looking at a commit where we didn't fix the line wrapping.

Anyways, I didn't realize there were substitutions done using `_FLAGS_REPLACE`. Do you think we should move those here to use `sed` or move these `sed` commands there to use `replace-word-pairs.py`. FWIW, I like the `sed` commands more because I think it's slightly more readable.

mingwandroid:
If i feel better tomorrow ill look into using the new replacements stuff for that part.

mingwandroid:
If you prefer sed for now then ok. I don't like the bsd vs gnu sed differences though and partly that's why i implemented the new replacements stuff which is also a lot more powerful.

mingwandroid:
One problem with the sed approach is the line splitting is done in python to the syscondif files. If we were to percent that then id be happier with sed  or the way

isuruf:
Line splitting issue was fixed by https://github.com/conda-forge/python-feedstock/pull/400/files AFAIR. I'll check

Issue #450, opened 2021-02-19
CVE-2021-3177


junpuf:
- [x] I read [the conda-forge documentation](https://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge) and could not find the solution for my problem there.

Issue: this issue pertains to [CVE-2021-3177](https://nvd.nist.gov/vuln/detail/CVE-2021-3177). I want to understand whether conda-forge is aware of this issue and is preparing updated Python releases. If so, I need to know when it will be released, and for which platforms.

Refer to this [doc](https://python-security.readthedocs.io/vuln/ctypes-buffer-overflow-pycarg_repr.html) for more detail about the fix.

<br/>
Environment (<code>conda list</code>):
<details>

```
$ conda list

```
</details>

<br/>
Details about  <code>conda</code> and system ( <code>conda info</code> ):
<details>

```
$ conda info

```
</details>

jakirkham:
Please raise Python bugs on the Python issue tracker ( https://bugs.python.org ). Thanks!

junpuf:
#449

jakirkham:
Gotcha. So this was already fixed. We are just waiting on upstream releases. Yes 3.7.10 shouldn't have the issue. Also 3.6.13 looks like it should have the fix. Though it doesn't look like 3.8 has seen a patch release yet. Could backport the patch and apply it though if that is straightforward.

junpuf:
@jakirkham I'm working on a PR for 3.6.13 at the moment. It looks that #449 is doing the 3.7.10

junpuf:
@jakirkham #451 CI is green, could you please review and approve, thank you.

junpuf:
@isuruf

jakirkham:
Thanks for the PRs. Have merged the 3.6 one. Had a question about the 3.7 one. Were you planning to submit a PR for 3.8? We can just apply the patch ourselves since they haven't released yet

junpuf:
@jakirkham thanks for the approval in #451.

For #452 I have first tried without the 0012 patch and run CI trying to understand if 0012 is the problem that caused the failure, and the CI was green, that confirmed my guess that the 0012 patch is the problem here.

I'll need some help to understand what tool should be used to refresh the 0012 patch in 3.7 branch.

junpuf:
[...]

PR #452, opened 2021-02-19
uptick version to 3.7.10

junpuf:
Checklist
~~* [ ] Bumped the build number (if the version is unchanged)~~

#450

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -70,7 +70,7 @@ source:
     sha256: de3c87b26a80e789986d8e6950c6304175d3829afe9c6c7211eb7257266ab0ac         # [win]
 
 build:
-  number: 0
+  number: 100
   # Windows has issues updating python if conda is using files itself.
   # Copy rather than link.
   no_link:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -29,7 +29,7 @@ source:
       - patches/0009-runtime_library_dir_option-Use-1st-word-of-CC-as-com.patch
       - patches/0010-Win32-Do-not-download-externals.patch
       - patches/0011-Add-support-for-_CONDA_PYTHON_SYSCONFIGDATA_NAME-if-.patch
-      #- patches/0012-Fix-find_library-so-that-it-looks-in-sys.prefix-lib-.patch
+      - patches/0012-Fix-find_library-so-that-it-looks-in-sys.prefix-lib-.patch
       - patches/0013-Disable-new-dtags-in-unixccompiler.py.patch
       - patches/0014-Fix-cross-compilation-on-Debian-based-distros.patch
       - patches/0015-Disable-registry-lookup-unless-CONDA_PY_ALLOW_REG_PA.patch
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set base_version = "3.7.9" %}
+{% set base_version = "3.7.10" %}
 {% set version = base_version %}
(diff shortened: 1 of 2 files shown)


junpuf:

recipe/meta.yaml
```diff
@@ -29,7 +29,7 @@ source:
       - patches/0009-runtime_library_dir_option-Use-1st-word-of-CC-as-com.patch
       - patches/0010-Win32-Do-not-download-externals.patch
       - patches/0011-Add-support-for-_CONDA_PYTHON_SYSCONFIGDATA_NAME-if-.patch
-      - patches/0012-Fix-find_library-so-that-it-looks-in-sys.prefix-lib-.patch
+      #- patches/0012-Fix-find_library-so-that-it-looks-in-sys.prefix-lib-.patch
```
jakirkham, line 32: Why is this patch commented?

jakirkham:
cc @mingwandroid (for vis)

recipe/meta.yaml
junpuf: this patch needs some fix, i first removed it to try it out, and CI passed. Now I'm trying to see how to fix the 0012 patch. I think some lines needs to be updated.

junpuf: looking at the 0012 patch file, it seems to be generated file. @jakirkham is there anything i should read about to fix this?

jakirkham: It's probably something simple like more code was added to the file so the line numbers are off. Basically this if just trying to prioritize to search for libraries to load in the Conda environment. Though I could be wrong about this

This was Ray's patch. Have already cc'd him. Would give him a chance to respond

recipe/patches/0012-Fix-find_library-so-that-it-looks-in-sys.prefix-lib-.patch
```diff
@@ -54,7 +54,7 @@ index 97973bc..1340ad7 100644
          def find_library(name):
              # See issue #9998
 -            return _findSoname_ldconfig(name) or \
--                   _get_soname(_findLib_gcc(name) or _findLib_ld(name))
+-                   _get_soname(_findLib_gcc(name)) or _get_soname(_findLib_ld(name))
```
junpuf, line 57: this may have some change, I'll figure this out.

isuruf: Have a look at 3.8 branch's patch. That might help. https://github.com/conda-forge/python-feedstock/blob/3.8/recipe/patches/0013-Fix-find_library-so-that-it-looks-in-sys.prefix-lib-.patch

junpuf: Yeah, that confirmed it. I think this is fine.

recipe/meta.yaml
```diff
@@ -70,7 +70,7 @@ source:
     sha256: de3c87b26a80e789986d8e6950c6304175d3829afe9c6c7211eb7257266ab0ac         # [win]
 
 build:
-  number: 100
+  number: 0
```
isuruf, line 73:
```suggestion
  number: 100
```

jakirkham: Why do we need this build number Isuru?

isuruf: To fix some issues where pypy is prioritised.

jakirkham: Ah ok. Do we need this for Python 3.6 as well?

PR #481, opened 2021-08-16
Update to 3.10.0rc1

isuruf:
- [x] remove upstreamed patch Fix-off-by-one-error-in-_winapi_WaitForMultipleObjec.patch
- [x] remove unnecessary patch Darwin-Add-arch-x86_64-to-UnicCCompiler-ld_args.patch
- [x] remove Add-support-for-_CONDA_PYTHON_SYSCONFIGDATA_NAME-if-.patch
- [x] unvendor libffi
- [x] unvendor bzip2
- [ ] unvendor xz
- [ ] unvendor zlib
- [ ] unvendor tcl/tk


recipe/conda_build_config.yaml
```diff
@@ -1,5 +1,5 @@
 python:
```
mbargull, line 1:
```suggestion
# Workaround: conda-build applies its default `python: {min_pin: x.x, max_pin: x.x}` after it
# resolved `{{ pin_subpackage('python', exact=True) }}`. Override this with "x.x.x.x.x" pins.
pin_run_as_build:
  python:
    min_pin: x.x.x.x.x
    max_pin: x.x.x.x.x
python:
```

mbargull: NB: I did not yet open an issue or draft PR for `conda-build` to resolve this upstream.

mbargull: (We actually need this (or the appropriate fix in `conda-build`) for non-RC builds too -- it just went unnoticed in the release versions.)

isuruf: That doesn't work either. `3.10.0.0.0 > 3.10.0.rc1`. We really need the conda-build PR I linked

mbargull: I tried it locally and it uses the version from the subpackage, i.e., this resolves to `python >=3.10.0rc.1,<3.10.0rc.2.0a0 h49503c6_0_cpython` which is okay-ish.

isuruf:
```yaml
$ cat recipe/meta.yaml 
package:
  name: foo
  version: 3.10.0.rc1

outputs:
  - name: foo
  - name: bar
    requirements:
      host:
        - foo
      run:
        - foo
    test:
      commands:
        - echo "hello world"

$ cat recipe/conda_build_config.yaml 
pin_run_as_build:
  foo:
    min_pin: x.x.x.x.x
    max_pin: x.x.x.x.x

foo:
  - 3.10
```

gives me `foo >=3.10,<3.11.0a0`

mariusvniekerk: we could just lie about it and call it 3.10.0.dev1.rc1 ?

isuruf: How would that help? `3.10.0.dev1.rc1 < 3.10.0` is still True

mbargull: @isuruf, oh fun. If you use `python` instead of `foo` you get the other behavior I described ("yay" for special case handling -_- ...). (And you get a stack trace if you use `pin_subpackage` with `3.10.0.rc1` instead of `3.10.0rc1`...)

recipe/meta.yaml
```diff
@@ -47,69 +47,58 @@ source:
...
   # TODO :: Depend on our own packages for these:
   - url: https://github.com/python/cpython-source-deps/archive/xz-5.2.2.zip  # [win]
     folder: externals/xz-5.2.2  # [win]
     sha256: 02b6d6f1e0369940e7e1f0d324992e069d09f5f3d32d86e2d22cd8b82ac3fb3f  # [win]
   # Not used as of 3.7b3 (tcltk from cpython-bin- is), but kept for archival and license reasons.
-  - url: https://github.com/python/cpython-source-deps/archive/tcl-core-8.6.9.0.zip  # [win]
-    folder: externals/tcl-core-8.6.9.0  # [win]
-    sha256: ea9e724536ec13b476ebbd47948303b92fdd90a33aa94f30fcc2f5c3a899f69e  # [win]
+  - url: https://github.com/python/cpython-source-deps/archive/tcl-core-8.6.11.0.zip  # [win]
+    folder: externals/tcl-core-8.6.11.0  # [win]
```
h-vetinari, line 87: The comment above sounds outdated (same below).

recipe/meta.yaml
```diff
@@ -197,15 +191,15 @@ outputs:
...
+        - bzip2
         - sqlite
         - xz  # [not win]
         - zlib  # [not win]
         - openssl
         - readline  # [not win]
         - tk  # [not win]
         - ncurses  # [unix]
-        - libffi   # [not (win or (osx and arm64))]
+        - libffi   # [not (osx and arm64)]
```
katietz, line 189: With libffi patch on Anaconda for arm64, it should be possible to use even for arm64 libffi package, instead of system one.  Major fix there is that the ffi_find_closure_for_code_np adjustment (Apple renamed the ffi_find_closure_for_code function)

katietz (approved):

recipe/meta.yaml
isuruf: Thanks @katietz for the info. Can you send a link to the patch?

katietz: Sure, it is part of this 3.3 patch for arm64.  We work right now on porting patch to 3.4 too. https://github.com/AnacondaRecipes/libffi-feedstock/blob/master/recipe/0003-libffi-3.3-arm64.patch
I used different sourceball and modified directly configuration stuff on libffi.  Sadly the apple-distribution python script has no support for arm64 for now ... therefore a reconfigure of autotool stuff would break it

katietz: see generate-darwin-source-and-headers.py file in libffi's root directory in sourceball.  It is neither adjusted in 3.3 for arm64, nor for osx64.  But it is required to setup the header-dependencies proper.  So this patch actually did this in a manual way ...

isuruf: Thanks. I've added a much smaller patch to 3.4. See the first two patches at https://github.com/conda-forge/libffi-feedstock/tree/master/recipe

recipe/build_base.sh
```diff
@@ -195,6 +191,7 @@ fi
 if [[ ${target_platform} == osx-64 ]]; then
   export MACHDEP=darwin
   export ac_sys_system=Darwin
+  # Don't set this as configure will not define _XOPEN_SOURCE_EXTENDED
   export ac_sys_release=13.4.0
```
h-vetinari, line 195: Isn't this comment now outdated?

recipe/patches/0015-Add-CondaEcosystemModifyDllSearchPath.patch
```diff
@@ -1,7 +1,51 @@
-From 9eb3eea71dcbe3c9c2649169bd35614998b1c650 Mon Sep 17 00:00:00 2001
+From 5bb1915898462b89ea9ed9791e26a2e7e3c63fd2 Mon Sep 17 00:00:00 2001
 From: Ray Donnelly <mingw.android@gmail.com>
 Date: Tue, 24 Dec 2019 18:37:17 +0100
-Subject: [PATCH 20/30] Add CondaEcosystemModifyDllSearchPath()
+Subject: [PATCH] Add CondaEcosystemModifyDllSearchPath()
+
+There are 3 depending on CONDA_DLL_SEARCH_MODIFICATION env variable
```
h-vetinari, line 6:
```suggestion
There are 3 modes depending on CONDA_DLL_SEARCH_MODIFICATION env variable
```

h-vetinari (commented):
Really cool! I especially like that much of the libs finally get unvendored also on windows.

recipe/patches/0015-Add-CondaEcosystemModifyDllSearchPath.patch
```diff
@@ -1,7 +1,48 @@
...
+    - The directory that contains the DLL
+    - Application directory
+    - C:\Windows
+    - Current working directory
+    - Above 5 directories if not in PATH
+    - PATH
+    - Directories added with os.add_dll_directory
+    - Old PATH entries (Deficiency in current patch)
+    - Old working directories (Deficiency in current patch)
+    - C:\Windows\System32
```
isuruf, line 42: @conda-forge/python, I've kept the `CONDA_DLL_SEARCH_MODIFICATION=1` code path as before in 3.8 and 3.9.
`CONDA_DLL_SEARCH_MODIFICATION` unset is changed now from 3.8 and 3.9.

In python3.8 upstream made the decision to change the DLL search order so that PATH is not looked at and instead asking library authors to rely on `os.add_dll_directory`. This broke conda and we reverted the upstream work. This came with the caveat that `os.add_dll_directory` doesn't work in conda python. See https://github.com/conda/conda/issues/10897. Although I wasn't in the conversation to revert upstream work to support I think it made sense at the time as otherwise it would have meant changing hundreds of packages.

I think with python3.10, we should go with upstream decision while making sure conda dependencies work. Issue with making PATH not visible is that, dependencies of python modules like `_ssl` (and python extensions) are not found as they are in a different library `<prefix>/Library/bin`. To fix this I add `<prefix>/Library/bin` and other conda paths using `os.add_dll_directory` when the python interpreter is initialized. There are 2 unintended advantages with this,
1. Running python.exe and importing python extensions without activating python environment works.
2. C:\Windows\System32 has the lowest priority. (Previously we needed to use `CONDA_DLL_SEARCH_MODIFICATION=1` to make System32 have the lowest priority in python 3.6-3.9. That has lots of other issues too.)

In order to keep the 1st advantage, I had to keep the vendored zlib module as python on windows doesn't have a separate zlib extension, but is embedded into the interpreter as a builtin module. We can either,
1. Keep zlib vendored in
2. Copy `<prefix>/Library/bin/zlib.dll` to `<prefix>` in zlib conda package (85 KB file).
I'm fine with either option.

@conda-forge/python, @conda-forge/core, @chenghlee, @jezdez, @katietz, thoughts on the change of DLL search order to not look in PATH and also on how to proceed with zlib?

recipe/patches/0015-Add-CondaEcosystemModifyDllSearchPath.patch
```diff
@@ -1,7 +1,48 @@
...
+    os.add_dll_directory(join(sys.prefix, 'Library', 'bin'))
+    os.add_dll_directory(join(sys.prefix, 'Library', 'usr', 'bin'))
+    os.add_dll_directory(join(sys.prefix, 'Library', 'mingw-w64', 'bin'))
+
+  Search order
+    - The directory that contains the DLL
+    - Application directory
+    - Directories added with os.add_dll_directory
+    - Above 5 directories
+    - C:\Windows\System32
```
h-vetinari, line 24:
```suggestion
  Search order
    - The directory that contains the DLL
    - Application directory
    - Directories added with os.add_dll_directory
    - Above 3 directories
    - C:\Windows\System32
```

recipe/patches/0015-Add-CondaEcosystemModifyDllSearchPath.patch
```diff
@@ -1,7 +1,48 @@
...
+- CONDA_DLL_SEARCH_MODIFICATION=1
+
+  Search order is roughly,
+
+    - The directory that contains the DLL
+    - Application directory
+    - C:\Windows
+    - Current working directory
+    - Above 5 directories if not in PATH
+    - PATH
```
h-vetinari, line 38: Same issue as above, though more importantly: does the "if not in PATH" still make sense if that bullet actually comes before the PATH itself?

```suggestion
    - The directory that contains the DLL
    - Application directory
    - C:\Windows
    - Current working directory
    - PATH
```

recipe/patches/0015-Add-CondaEcosystemModifyDllSearchPath.patch
isuruf: It's not the 3 directories directly above. The 5 directories mentioned earlier in the comment

recipe/patches/0015-Add-CondaEcosystemModifyDllSearchPath.patch
isuruf: That's not the correct order.

h-vetinari: Yeah, my confusion was the same as in the above comment. Would similarly suggest:
`Above 5 directories if not in PATH` -> `The 5 (as if added by "os.add_dll_directory") directories above, if not in PATH`

katietz (approved):
Thanks, this documentation adjustments clarify what was meant by "Above 5 directories".

recipe/patches/0015-Add-CondaEcosystemModifyDllSearchPath.patch
xhochy: Regarding `zlib`, I would go with either 1 or 2. If we go for 2, we should do the copy as part of the `zlib` package. Personally, I have a slight preference for 2 but I'm happy with 1 and we can do 2 as a follow-up PR.

mbargull: Re `zlib`, putting `zlib.dll` in the root for `conda-forge::zlib` isn't necessarily sufficient since we cannot assume users to have "clean" environments, i.e., it can happen that they have `defaults::zlib` installed (or install it) (I know this is not something we want, but esp. packages like `python` are installed by many unaware/inexperienced users). Less likely (probably not highly likely due to `zlib` having few dependencies), but still possible, is also that people get an older `zlib` build installed since `conda` does not prevent this.

recipe/meta.yaml
```diff
@@ -199,21 +170,21 @@ outputs:
...
+        - zlib
         - openssl
         - readline  # [not win]
-        - tk  # [not win]
+        - tk
         - ncurses  # [unix]
-        - libffi   # [not (win or (osx and arm64))]
+        - libffi
         - ld_impl_{{ target_platform }} >=2.36.1  # [linux]
+        - libuuid  # [unix]
```
mbargull, line 183: Do we want `- libuuid  # [linux]` here and link to `libSystem.dylib` (or wherever the libuuid symbols live on macOS)? This is regarding https://github.com/conda-forge/libuuid-feedstock/issues/16 .
(I'm only aware of the above issue but lack knowledge on the details so can't say which way is better one, sorry.)

recipe/patches/0015-Add-CondaEcosystemModifyDllSearchPath.patch
isuruf: @mbargull, with https://github.com/conda-forge/zlib-feedstock/pull/52, this shouldn't be a problem as the new package `libzlib` will always have `zlib.dll` in the root. (Assuming defaults will use our recipe to have `libzlib`.)

recipe/meta.yaml
mbargull: Going off of https://github.com/regro/cf-scripts/issues/417 and the linked issues, IIUC, at least one problem is that if `libuuid` is pulled into an environment, then `include/uuid/uuid.h` could conflict with the one provided by the macOS SDK. I.e, this would more be a downstream issue than for `python` itself.
Maybe @chrisburr know more and could comment if pulling in `libuuid` via `python` could lead to problems?

recipe/patches/0015-Add-CondaEcosystemModifyDllSearchPath.patch
mbargull: Hehe, yep, nice opportunity to split ;).

recipe/meta.yaml
isuruf: > at least one problem is that if libuuid is pulled into an environment, then include/uuid/uuid.h could conflict with the one provided by the macOS SDK.

How would it conflict though?

isuruf: nvm, saw the conversation in root

jakirkham:
Looks like the final release just dropped

https://github.com/conda-forge/python-feedstock/pull/509

h-vetinari:
This is amazing work 🥇

chrisburr (approved):
Amazing work! 🍰

recipe/build_base.bat
```diff
@@ -82,12 +82,6 @@ if errorlevel 1 exit 1
 mkdir %PREFIX%\DLLs
 xcopy /s /y %SRC_DIR%\PCBuild\%BUILD_PATH%\*.pyd %PREFIX%\DLLs\
 if errorlevel 1 exit 1
-copy /Y %SRC_DIR%\PCbuild\%BUILD_PATH%\tcl86t.dll %PREFIX%\DLLs\
```
mariusvniekerk, line 85: So much nicer

ocefpaf (approved):

recipe/patches/0027-Set-ssltag-to-3.patch
```diff
@@ -0,0 +1,39 @@
+From eb3e4e0dbd5a925825cd2fb989033d8b1860cfc0 Mon Sep 17 00:00:00 2001
+From: Chris Burr <chrisburr@users.noreply.github.com>
+Date: Tue, 28 Sep 2021 12:40:44 +0200
+Subject: [PATCH 29/29] Set ssltag to 3
```
isuruf, line 4: What's the error you get when this happens?

chrisburr: On Windows it tries to link to the wrong shared library name (see the CI in the Python 3.9 OpenSSL PR)

PR #507, opened 2021-09-26
depend on libnsl

izahn:
Checklist

This is an attempt to fix the `libnsl` dependency issue reported and discussed at https://github.com/conda-forge/rasterio-feedstock/issues/220 and https://github.com/conda-forge/cmake-feedstock/issues/69. It needs https://github.com/conda-forge/staged-recipes/pull/16278 to be merged first.


izahn:

xylar:
As discussed in https://github.com/conda-forge/rasterio-feedstock/issues/220, the original poster is stating that libnsl shouldn't be needed and is suggesting that it be explicitly dropped from the build.  So it seems worth discussing if that is a possibility instead of adding it explicitly.

isuruf:
> libnsl shouldn't be needed and is suggesting that it be explicitly dropped from the build. 

What? that doesn't make sense to me at all.

xylar:
@MichaelTiemannOSC, would you like to clarify why that would be your preferred way forward?

izahn:
As an alternative to this PR it may be possible to remove the nis module as described at https://python-build-standalone.readthedocs.io/en/latest/technotes.html#libnsl-nis-module .

MichaelTiemannOSC:
Thank you for inviting me into this discussion.  For context, I entered the dev world in the days of RCS and early days of CVS.  I use git--as everybody does these days--and am familiar with libraries and dependencies, which challenging in the late 1980s and exponentially more challenging today.  I came into this python-feedstock discussion via the rasterio-feedstock repo as my original problem had to do with a dependency problem with rasterio.

My understanding, which could be completely wrong, is that the PyPi package makes use of a cpython/cython component for the _base functionality in rasterio that doesn't depend on libnsl.so.1.  My suggestion for a preferred way forward is to either depend on the much-less-deprecated libnsl2 (if there's no two ways around using that library) or to wean away from libnsl.so.1.  In either case, this will require some changes in rasterio, which I'm happy to propose to them.  But I followed their (website) guidance that the first step to addressing packaging problems was to visit the feedstock workstream, which I did.

The response from the feedstock team, predictably, was to create a solution that did not require touching any rasterio code, by making the deprecated library libnsl.so.1 a stand-alone thing that could be dependent upon.  I think that's not a great approach as it makes anachronistic code live forever.  But it is an approach.

With all of the above, let me approach the rasterio maintainers and see if they are game to fix things in 1.2.9 or a 1.3.x version.  Thanks!

izahn:
> My suggestion for a preferred way forward is to either depend on the much-less-deprecated libnsl2 (if there's no two ways around using that library) or to wean away from libnsl.so.1

This PR is exactly an attempt to depend on `libnsl2` by packaging it and adding it to the `host` requirements here.

izahn:
@conda-forge-admin please restart ci

izahn:

izahn:
> Can you send PRs to `master (3.10), 3.8, 3.7` branches?

submitted in:

- https://github.com/conda-forge/python-feedstock/pull/512 (3.10)
- https://github.com/conda-forge/python-feedstock/pull/513 (3.8)
- https://github.com/conda-forge/python-feedstock/pull/514 (3.7)

Issue #516, opened 2021-10-13
python 3.7 + openssl 3 on windows fails with ssl.SSLError: unknown error (_ssl.c:4034)


akrherz:
Issue: I don't believe python 3.7 + openssl 3 is supported upstream, conda-forge python 3.7 + openssl 3 fails on windows.

<br/>
Environment (<code>conda list</code>):
<details>

```
$ conda list
# packages in environment at C:\Users\akrherz\Miniconda3\envs\py37:
#
# Name                    Version                   Build  Channel
ca-certificates           2021.10.8            h5b45459_0    conda-forge
openssl                   3.0.0                h8ffe710_1    conda-forge
pip                       21.3               pyhd8ed1ab_0    conda-forge
python                    3.7.10          h900ac77_102_cpython    conda-forge
python_abi                3.7                     2_cp37m    conda-forge
setuptools                58.2.0           py37h03978a9_0    conda-forge
sqlite                    3.36.0               h8ffe710_2    conda-forge
ucrt                      10.0.20348.0         h57928b3_0    conda-forge
vc                        14.2                 hb210afc_5    conda-forge
vs2015_runtime            14.29.30037          h902a5da_5    conda-forge
wheel                     0.37.0             pyhd8ed1ab_1    conda-forge
```
</details>

<br/>
Details about  <code>conda</code> and system ( <code>conda info</code> ):
<details>

```
$ conda info
     active environment : py37
    active env location : C:\Users\akrherz\Miniconda3\envs\py37
            shell level : 2
       user config file : C:\Users\akrherz\.condarc
 populated config files : C:\Users\akrherz\.condarc
          conda version : 4.10.3
    conda-build version : not installed
         python version : 3.9.5.final.0
       virtual packages : __win=0=0
                          __archspec=1=x86_64
       base environment : C:\Users\akrherz\Miniconda3  (writable)
      conda av data dir : C:\Users\akrherz\Miniconda3\etc\conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/win-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/win-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/msys2/win-64
[...]

PR #517, opened 2021-10-13
Fix Openssl3 + py37 on windows failure

akrherz:
Checklist

Adds a test showing the openssl3 + python3.7 failure.  Will iterate on a patch to then fix this.

Fixes https://github.com/conda-forge/python-feedstock/issues/516

### DIFF
--- a/recipe/patches/0022-Openssl3-load-verify-error.patch
+++ b/recipe/patches/0022-Openssl3-load-verify-error.patch
@@ -1,3 +1,12 @@
+From 82b6c0909aae423d9c8f4ff7d0e8df16106dbe28 Mon Sep 17 00:00:00 2001
+From: Christian Heimes <christian@python.org>
+Date: Fri, 23 Apr 2021 16:37:09 +0200
+Subject: [PATCH] [3.8] bpo-43920: Make load_verify_locations(cadata) error
+ message consistent (GH-25554) (GH-25556)
+
+Co-authored-by: akrherz <akrherz@iastate.edu>.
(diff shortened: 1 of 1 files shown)


akrherz:
Great, the failure reproduced only on windows + openssl3
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "D:\bld\python_1634123290171\_test_env\lib\ssl.py", line 589, in create_default_context
    context.load_default_certs(purpose)
  File "D:\bld\python_1634123290171\_test_env\lib\ssl.py", line 490, in load_default_certs
    self._load_windows_store_certs(storename, purpose)
  File "D:\bld\python_1634123290171\_test_env\lib\ssl.py", line 482, in _load_windows_store_certs
    self.load_verify_locations(cadata=certs)
ssl.SSLError: unknown error (_ssl.c:4034)
```

Patch iteration coming in a bit

akrherz:
OK, this should be ready to go.  My understanding of why this patch fixes the issue is that the ASN1 reader was incorrectly erroring when it found an unexpected EOF on the `bytearray`, but it should have been OK.  This code path is unique to windows.

jakirkham:
Woot! 🎉 Thanks @akrherz! 😄 

Would it make sense to add the `urlopen` example from issue ( https://github.com/conda-forge/python-feedstock/issues/516 ) as a test as well?

akrherz:
> Would it make sense to add the `urlopen` example from issue ( #516 ) as a test as well?

I don't believe so as that would introduce a network availability requirement for the test to succeed.  I added a test that exercises the same failing codepath.

Issue #522, opened 2021-11-11
Python 3.10.0 crashes on MacOS x86_64


akrherz:
- [x] I read [the conda-forge documentation](https://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge) and could not find the solution for my problem there.

Issue: Two conda-forge repos having strange crashes within CI on MacOS with python 3.10.0

Note: I am not a mac user, but am just attempting to help conda-forge repos get green CI builds :)

Examples of CI crashes: conda-forge/magics-feedstock#106 and conda-forge/shapely-feedstock#92

I do have an old mac that I am able to reproduce the shapely crashes and have gotten some iterative feedback here along with `lldb` output:  Toblerity/Shapely#1227

Does anybody here have interest / mac experience to help figure out what is wrong?  Thanks!

dopplershift:
I'll add to this (because I think they are related) the `Abort trap: 6` (the same as the error from Magics) that we're seeing in conda-forge/rtree-feedstock#41 and conda-forge/python-libarchive-c-feedstock#29.

dopplershift:
So something is definitely broken about ctypes in the conda-forge build. Shapely builds and runs tests fine on my system with Homebrew Python 3.10; more importantly, if I copy `_ctypes.cpython-310-darwin.so` from the Homebrew Python 3.10 into my conda-forge one, Shapely's tests run without crashing.

dopplershift:
I took the Python recipe and did a local build where I removed `libffi` from `host`. Running `conda build` for all four of rtree, python-libarchive-c, shapely, and magics worked fine (tests completed) using my local build of Python.

Is there any particular reason to avoid the macOS copy of libffi? [This Python issue](https://bugs.python.org/issue42688) would seem to indicate that using the system copy for macos is the norm.

xhochy:
Pinging @isuruf and @erykoff here as they have had to experience a lot of issues with `libffi` on macOS.

One important difference to note here is that these issues only seem to happen on `osx-64` where we historically have build with conda-forge's `libffi` already for the last releases. Only `osx-arm64` now also uses the c-f `libffi`, too.

jakirkham:
Actually this raises a question that is unclear (at least to me). Is this occurring with macOS ARM or x86?

erykoff:
[...]

PR #523, opened 2021-11-19
Add segfault test on osx-64+libffi=3.4 (bpo-44556)

mbargull:
refs:
  - https://bugs.python.org/issue44556
  - https://github.com/conda-forge/python-feedstock/issues/522
  - https://github.com/conda-forge/llvmlite-feedstock/pull/58#issuecomment-973829079

Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #524, opened 2021-11-19
create manual new python build against libffi=3.3

akrherz:
Checklist
* [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

This is perhaps a lame attempt at creating a python3.10 + libffi build combination that other packages could pin against to avoid bugs perhaps with the libffi=3.4 package.  ie #522 

Totally feel free to reject this forcibly, just attempt to be helpful :)

I purposefully did not re-render.

### DIFF
--- a/.ci_support/linux_64_openssl1.1.1.yaml
+++ b/.ci_support/linux_64_openssl1.1.1.yaml
@@ -17,7 +17,7 @@ cxx_compiler_version:
 docker_image:
 - quay.io/condaforge/linux-anvil-comp7
 libffi:
-- '3.4'
+- '3.3'
 libuuid:
 - '2'
 ncurses:
--- a/.ci_support/linux_64_openssl3.yaml
+++ b/.ci_support/linux_64_openssl3.yaml
@@ -17,7 +17,7 @@ cxx_compiler_version:
 docker_image:
 - quay.io/condaforge/linux-anvil-comp7
 libffi:
-- '3.4'
+- '3.3'
 libuuid:
 - '2'
 ncurses:
--- a/.ci_support/linux_aarch64_openssl1.1.1.yaml
+++ b/.ci_support/linux_aarch64_openssl1.1.1.yaml
@@ -21,7 +21,7 @@ cxx_compiler_version:
 docker_image:
 - quay.io/condaforge/linux-anvil-aarch64
 libffi:
-- '3.4'
+- '3.3'
 libuuid:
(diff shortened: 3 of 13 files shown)


jakirkham:
Thanks Daryl! 😄

As @erykoff came up with a workable solution ( https://github.com/conda-forge/python-feedstock/issues/522#issuecomment-974521764 ), would propose we do that instead. Also would suggest we include the test @mbargull came up with ( https://github.com/conda-forge/python-feedstock/pull/523 )

cc @beckermr (for vis)

beckermr:
My understanding is that 3.3 working is an accident. If apple ever changes what their system version is, things would break again. So @erykoff 's solution is preferred.

akrherz:
> My understanding is that 3.3 working is an accident. If apple ever changes what their system version is, things would break again.

I don't agree.  If the bug is that python got compiled against system libffi, but linked at runtime against conda-forge libffi.  The fact that this build hard codes this runtime link against 3.3 would prevent any future system updates from causing issues.

> So @erykoff 's solution is preferred.

Totally agreed, just sort of nice to have an immediate green alternative :)

erykoff:
I think that Matt's point is not that the build here would break, but that this way of building would break if run on a system that had libffi 3.4.2.

I think it should be very easy to instead to `export PKG_CONFIG=$BUILD_PREFIX/bin/pkg-config` in build_base.sh and that would get this going.

akrherz:
I don't mean to argue or make ignorant noise, but will also note that this

> $ python -c "from ctypes import CFUNCTYPE; CFUNCTYPE(None)(id)"

does not reproduce for me locally on 10.13.6

erykoff:
Yeah, I couldn't get that to crash either.  I was basing my investigations on the crashes in the python ctypes tests.  And putting in debugging statements led to a whole bunch of compiler warnings which led me to the include problem.

jakirkham:
If you would like to add an additional test, that would be welcome :) Ideally it would be something self-contained to `python` (so no other dependencies)

isuruf:
Superceded by #525

akrherz:
Thanks @isuruf , I was about to mark this as a draft :)  It would have been nice to see what CI did though...

PR #527, opened 2021-12-02
Bump to 3.6.15

teake:
Checklist

The driver for this PR is https://github.com/python/cpython/pull/27251 (without which Jupyter Lab doesn't run on macOS Big Sur), which has been backported to 3.6.15 according to https://github.com/jupyterlab/jupyterlab/issues/9410#issuecomment-955705925.


teake:

teake:
> Hmm, I had hoped a simple version bump would suffice, but alas. Looks like the `nis` package doesn't compile on linux64 and aarch64. Any suggestions for fixing this? I must admit I'm a little out of my depth here ...

Adding the `libsnl` dependency seems to have worked :).

recipe/meta.yaml
```diff
@@ -124,6 +124,7 @@ requirements:
     - tk                                 # [unix]
     - ncurses                            # [unix]
     - libffi                             # [unix]
+    - libnsl                             # [linux]
```
ocefpaf, line 127: Looks good to me but I'd be more confident if @isuruf could take a quick look before we merge.

isuruf: That's fine, but note that we don't support python 3.6 anymore.

ocefpaf: Yeah. I still think having a new py36 build is good for bugfixes. But building new packages for py36 is a no-no due to CI usage.

PR #529, opened 2021-12-10
Update to python v3.9.9

jnooree:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

PR #530, opened 2021-12-17
python 3.9.9

xhochy:
Fixes #529 

Checklist


isuruf:
Should we backport the CONDA_DLL_SEARCH_MODIFICATION_ENABLE patch from 3.10 ?

xhochy:
> Should we backport the CONDA_DLL_SEARCH_MODIFICATION_ENABLE patch from 3.10 ?

This means to ship the updated version of https://github.com/conda-forge/python-feedstock/blob/2dc00628de43088e6c1c72bcb2bfe74c200d8b02/recipe/patches/0014-Add-CondaEcosystemModifyDllSearchPath.patch#L1 in this branch instead of the current patch we have for that?

xhochy:
I would be in favor of doing this and would do the work but I'm not sure what all I need to backport.

xhochy:
Should we switch `aarch64` also to emulated here as we did for 3.10?

isuruf:
> This means to ship the updated version of in this branch instead of the current patch we have for that?

Yes.

> I would be in favor of doing this and would do the work but I'm not sure what all I need to backport.

Thanks for volunteering.

> Should we switch aarch64 also to emulated here as we did for 3.10?

Yes

isuruf:
Maybe it's better to use all of the changes of 3.10 and port them here. Let me know if you can do that. If not, I'll try to do it.

xhochy:
> Maybe it's better to use all of the changes of 3.10 and port them here. Let me know if you can do that. If not, I'll try to do it.

I don't have a clear enough picture to do this. If you can list them, I can try.

isuruf:
Basically do https://github.com/conda-forge/python-feedstock/blob/master/recipe/patches/README.md with old=3.10.0 and new=3.9.9

Issue #575, opened 2022-08-09
Loading `libm` on macOS fails with Python 3.7


jakirkham:
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

```python
Python 3.7.12 | packaged by conda-forge | (default, Oct 26 2021, 05:59:23) 
[Clang 11.1.0 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from ctypes.util import find_library
>>> assert find_library("libm") is not None
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError
```

Note: This works with Python 3.8+. Maybe we are missing a patch with Python 3.7?

### Installed packages

```shell
# packages in environment at /Users/jkirkham/miniconda/envs/py37:
#
# Name                    Version                   Build  Channel
appnope                   0.1.3              pyhd8ed1ab_0    conda-forge
attrs                     22.1.0             pyh71513ae_1    conda-forge
backcall                  0.2.0              pyh9f0ad1d_0    conda-forge
backports                 1.0                        py_2    conda-forge
backports.functools_lru_cache 1.6.4              pyhd8ed1ab_0    conda-forge
blas                      2.115                  openblas    conda-forge
blas-devel                3.9.0           15_osx64_openblas    conda-forge
ca-certificates           2022.6.15            h033912b_0    conda-forge
cloudpickle               2.1.0              pyhd8ed1ab_0    conda-forge
cytoolz                   0.12.0           py37h994c40b_0    conda-forge
decorator                 5.1.1              pyhd8ed1ab_0    conda-forge
flake8                    5.0.4              pyhd8ed1ab_0    conda-forge
fsspec                    2022.7.1           pyhd8ed1ab_0    conda-forge
importlib-metadata        4.11.4           py37hf985489_0    conda-forge
importlib_metadata        4.11.4               hd8ed1ab_0    conda-forge
iniconfig                 1.1.1              pyh9f0ad1d_0    conda-forge
ipython                   7.33.0           py37hf985489_0    conda-forge
jedi                      0.18.1             pyhd8ed1ab_2    conda-forge
libblas                   3.9.0           15_osx64_openblas    conda-forge
libcblas                  3.9.0           15_osx64_openblas    conda-forge
libcxx                    14.0.6               hce7ea42_0    conda-forge
libffi                    3.4.2                h0d85af4_5    conda-forge
[...]

PR #577, opened 2022-08-23
3.11.0

isuruf:
- [x] Remove CONDA_DLL_SEARCH_MODIFICATION=1 codepath as it's not needed anymore


isuruf:
@conda-forge-admin, rerender

mariusvniekerk:
Looks like the downstreams tests are failing

cclauss:
Upgrade to rc2?  https://www.python.org/download/pre-releases/

recipe/meta.yaml
```diff
@@ -140,8 +139,6 @@ outputs:
         # Putting these here means they get emitted to build_env_setup.{sh,bat} meaning we can launch IDEs
         # after sourcing or calling that script without examine the contents of conda_build.{sh,bat} for
         # important env. vars.
-        - OPENSSL_DIR=%PREFIX%\Library    # [win]
```
katietz, line 143: This seems to be suspicious. For us, on anaconda, those lines seem still be required.  Why it works on your side without them?

recipe/conda_build_config.yaml
```diff
@@ -1,14 +1,8 @@
...
 numpy:
   - 1.16
-c_compiler:                    # [win]
-  - vs2017                     # [win]
-cxx_compiler:                  # [win]
-  - vs2017                     # [win]
-vc:                            # [win]
-  - 14.1                       # [win]
 MACOSX_SDK_VERSION:            # [osx and x86_64]
   - 11.0                       # [osx and x86_64]
```
katietz, line 8: Is this intended?  We can built it pretty nicely with 10.14 SDK on osx-64

recipe/meta.yaml
isuruf: You might not have the latest version of the patch https://github.com/conda-forge/python-feedstock/blob/071818b32099122e18735c6c7a40591250bca4c5/recipe/patches/0009-Unvendor-openssl.patch

recipe/conda_build_config.yaml
isuruf: Yes, it's intended. Building with 10.14 will give you a python interpreter that's broken on 11.0. For eg: https://github.com/conda-forge/python-feedstock/issues/575

recipe/meta.yaml
danpetry: I applied this and it works now, thanks

recipe/conda_build_config.yaml
danpetry: However, using 11.0 doesn't build for intel 64 macos here

danpetry: We tested Python 3.7-10 which was built with SDK v10.10 on OSX v12 and didn't get the assertion that jakirkham got in the above issue. We are going to build out with 10.4 and test, will keep you updated

danpetry: If you have any more thoughts please let me know

isuruf: > However, using 11.0 doesn't build for intel 64 macos here. 

I don't understand what you are saying here. Details please.

danpetry: If we use macOS SDK version 11.0, we can't build python 3.11 for the osx-64 target. (64-bit intel CPU Mac). However, for previous versions of Python (we're on 3.10.4 at the moment), we successfully used SDK version 11.0 to build for all targets (both ARM and Intel mac).

isuruf: What is the error when you use SDK 11.0?

danpetry: ```/var/folders/sy/f16zz6x50xz3113nwtb9bvq00000gp/T/abs_2092500kee/croots/recipe/python-split_1663247373378/work/Objects/longobject.c:5693:5: error: implicit declaration of function 'static_assert' is invalid in C99 [-Werror,-Wimplicit-function-declaration]```

The only change between a successful build, and a build with this failure, is changing the SDK version from 10.14 to 11.0.

The `static_assert` call in question was added somewhere between Python v3.10.4 and v3.11.

h-vetinari: > `implicit declaration of function 'static_assert' is invalid in C99 [-Werror,-Wimplicit-function-declaration]`

That's a consequence of upstream python [switching](https://github.com/python/cpython/pull/91733) to C11, which makes `static_assert` a keyword (where it might might have been e.g. a macro before; xref upstream [PR](https://github.com/python/cpython/pull/91730)).

Have you verified that you're compiling with `-std=c11`?

h-vetinari: That topic is also discussed quite extensively in https://github.com/python/cpython/issues/91731

chrisburr (approved):
I've made a 3.10 branch based on the current master and am about to re-target the existing PRs.

jakirkham:
Guess the next step is starting a migration, added a draft PR ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3573 ) to get this started. Feedback welcome 🙂

chrisburr:
I've also updated `python_abi` https://github.com/conda-forge/python_abi-feedstock/pull/16 as I'm fairly sure the bot isn't able to.

Issue #594, opened 2022-11-07
Are debug builds supported? 


nikhilweee:
### Comment:

I see a `--with-pydebug` option specified in the recipe folder. I wonder what's it for? 
https://github.com/conda-forge/python-feedstock/blob/018b46521412f51cf732a1e4e08e999cb160f6bd/recipe/build_base.sh#L51-L58
Me along with a few others have been on the lookout for a debug build (https://github.com/conda-forge/staged-recipes/issues/4812) and I wonder if this is the way to go?

jakirkham:
It's possible to build locally (as seen above), but we do not currently publish debug packages

nikhilweee:
Is the community interested in reviving #86? Although I don't have much experience writing conda recipes, I am eager to take a shot at it.

jakirkham:
Idk when I've asked other developers about whether a debug Python is useful, have heard they may want to make additional changes to the `python` build itself, use other debug libraries in the build, and/or build other downstream dependencies with changes. Given this, it has been unclear (at least to me) that there is one debug build of Python that would be helpful for these different use cases

Issue #73, opened 2016-10-02
Debug builds?


wesm:
Any idea how much work it would be to have a, say, `3.5-debug` Python version? Maybe this is already available somewhere.

tadeu:
@wesm Also take a look at https://github.com/conda-forge/staged-recipes/issues/1593 and https://github.com/conda-forge/staged-recipes/pull/961

dfroger:
Just created the pull request #86 about this

jakirkham:
This is a great idea that we should really revive (especially as the tooling has caught up a bit). Raised issue ( https://github.com/conda-forge/conda-forge.github.io/issues/544 ) to discuss how we do split packages and generate things like debug builds as products.

jakirkham:
@msarahan, do you know if any work has been done on this front in `defaults`? What recommendations do you and/or others have in terms of building and selecting between debug and non-debug builds of `python` both in downstream builds and `conda` installs? Are there options that `conda-build` 3 provides that would be helpful for this effort?

jakirkham:
It's worth noting as of Python 3.8 the interpreter includes all debug symbols. So a separate debug build is no longer needed for 3.8+.

ref: https://docs.python.org/3/whatsnew/3.8.html#debug-build-uses-the-same-abi-as-release-build

cc @conda-forge/core (for visibility)

jakirkham:
With some of the recent work on PyPy, this seems more doable. Raising issue ( https://github.com/conda-forge/python-feedstock/issues/73 ) to discuss.

mingwandroid:
I did some work for debug builds on Windows on Python 3.8. The ABI is not exactly the same. There is one symbol of difference ("CreateModule2" gets suffixed "Refcounted" or something like that).

I used this when I was looking into a Python 3.8 issue on Windows with LIEF. It builds a load of stuff with a huge amount of debug info using a very hacky batch file. To change what gets built, see:
If anyone has time, this is how to run it. You need my latest WIP conda-build branch too! 
```
conda activate
cd %TEMP%
mkdir debug-py-win
cd debug-py-win
git clone https://github.com/AnacondaRecipes/python-feedstock
git clone https://github.com/mingwandroid/conda-build -b 10-Add-build-id-pat-option
pushd conda-build
  python setup.py develop
popd
curl -SLO https://raw.githubusercontent.com/AnacondaRecipes/aggregate/master/conda_build_config-dbg_c-dbg_py_win-32.yaml
[...]

PR #596, opened 2022-11-11
Add a debug version of python

nikhilweee:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

## Background
There have been multiple conversations around supporting a debug build of python both in this repository (#73) and elsewhere (https://github.com/conda-forge/staged-recipes/issues/4812). There have also been previous attempts at this (#86) but there hasn't been any progress in the last 6 years. This PR is an attempt to revive those efforts.

## Details
I am fairly new to writing recipes for conda-forge. In fact, I have never written recipes before, so please be humble with all the help you can. What I propose is to have another output, `python-debug`, which passes the env vars `PY_INTERP_DEBUG=yes` and `DEBUG_PY=yes`. The build script `build_base.sh` is already equipped to handle these env vars and pass on `--with-pydebug` to the build process. 

## Benefits
* With this PR, users should be able to `conda install python-debug` to use a debug version of python.

## Concerns
* Adding an additional output requires more CI resources, something that we might want to conserve. 
* Adding another output makes `meta.yaml` even more messier.

Any feedback is much appreciated.

PR #597, opened 2022-11-13
Add a debug build of python 3.11

### LINKED ISSUE #598: @conda-forge-admin, please re-render
@conda-forge-admin, please re-render.

nikhilweee:
Checklist

There have been multiple conversations around supporting a debug build of python both in this repository (#73) and elsewhere (https://github.com/conda-forge/staged-recipes/issues/4812). There have also been previous attempts at this (#86) but there hasn't been any progress in the last 6 years. This PR is an attempt to revive those efforts.

Fixes https://github.com/conda-forge/python-feedstock/issues/594
Fixes https://github.com/conda-forge/python-feedstock/issues/73
Closes https://github.com/conda-forge/python-feedstock/pull/86
Fixes https://github.com/conda-forge/staged-recipes/issues/1593
Fixes https://github.com/conda-forge/conda-forge.github.io/issues/1017

Closes #598 #599 #600


nikhilweee:

nikhilweee:
@conda-forge-admin, please ping team

Hello team, I'm new to building packages for conda-forge but I'm really interested in a debug build of python. I spent the last couple days trying to figure out a way to include a debug build of python. Initially, in #596 I tried adding a separate output called `python-debug` but that didn't go so well because compiling `python-debug` would overwrite the build files for `python`. I couldn't find a way to use separate build folders so I gathered that it's a bad idea to build both packages simultaneously. 

In this PR, I basically changed the name of the output to `python-debug` and made sure that `--with-pydebug` is being passed on to `configure`. All goes well until conda runs tests. As you can see in the reference below, conda explicitly adds `python` as a dependency which causes all sorts of clobber warnings in the CI builds.  I am starting to run out of ideas and would really appreciate some help. Thanks!

https://github.com/conda/conda-build/blob/cc6c050ee3325e6957dbf60f74dbe79d78d26a8f/conda_build/metadata.py#L2365-L2369

nikhilweee:
Apologies for the double mention, didn't know that the bot will retrigger on edits.

jakirkham:
We may want to recheck this Windows ABI incompatibility issue that Ray had flagged ( https://github.com/conda-forge/python-feedstock/issues/73#issuecomment-605151779 )

nikhilweee:
Okay so after the last comment I figured that I shouldn't change the package name because that's how conda figures out not to install another version of python if this debug version is already installed. However, I did change the build string to include `dbg`. I believe now people can install a debug version using `conda install python=*=*dbg* -c conda-forge`

@isuruf I didn't intend to merge this into `main` because that would mean overwriting the non-debug version. I was under the impression that maybe we could maintain `{VER}-dbg` branches separately and update them less often but I'm open to other ideas. On another note, after almost a week of struggling, the linux builds are now passing and I feel good about my first PR.

jakirkham:
Think it would be better to build this with the `main` branch and just add extra jobs or iterate over builds in a loop (whichever is easiest). If we stick this in a different branch, it will likely fall out of date and become incompatible with the normal `python` build. Additionally sticking this in `main` makes it easier to backport this change to older `python` versions.

nikhilweee:
In that case I should add the flags used to differentiate between the two builds. I had removed it earlier.

jakirkham:
We could add [a variant]( https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#using-variants-with-the-conda-build-api ) to [`conda_build_config.yaml`]( https://github.com/conda-forge/python-feedstock/blob/018b46521412f51cf732a1e4e08e999cb160f6bd/recipe/conda_build_config.yaml ) perhaps [like so]( https://conda-forge.org/docs/maintainer/knowledge_base.html#build-matrices ). Once incorporated into the recipe (and [re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks )), this should generate the additional jobs

nikhilweee:
Thanks @jakirkham, that was very helpful. I added two variants in `conda_build_config.yaml`. Let's see how the tests go.

nikhilweee:

nikhilweee:

recipe/meta.yaml
```diff
@@ -136,6 +136,7 @@ outputs:
       script_env:
         - PY_INTERP_LINKAGE_NATURE={{ linkage_nature_env }}
         - PY_INTERP_DEBUG={{ py_interp_debug }}
+        - DEBUG_PY={{ py_interp_debug }}
```
jakirkham, line 139: Why not just use `PY_INTERP_DEBUG`?

nikhilweee: Because `build_static.sh` needs `DEBUG_PY` and that's why the latest CI builds failed.

nikhilweee: Also because `build_base.sh` uses `DEBUG_PY` to add `--with-pydebug` to `./configure`.

nikhilweee: I could change those scripts to just rely on `PY_INTERP_DEBUG` instead.

mbargull (dismissed):
We should ensure that the debug builds are only installed if explicitly requested.
Some possibilities:
1. Push to a `debug` label so people can `install conda-forge/label/debug::python`.
   (A separate `debug_python` label could also be possible if people see the need for it.)
2. Add a `run` requirement on some `debug`/`_debug` meta-package.
   (The `debug`/`_debug` package would then live in another channel/label (or have `track_feaures`).)
4. Add some `track_features: [debug]`.

Anaconda had some old 2.7 debug builds an used `features: [debug]` with an accompanying `debug` packages that had the corresponding `track_features`.
Nowadays we don't use `features` anymore but do still have `track_features` where needed.
The main cases for `track_features` are when we want certain variants/packages not installed by default but still have the possibility so seamlessly introduce them even implicitly (e.g. via dependencies).
Since we shouldn't have the need for that implicit behavior and `track_features` is a non-perfect workaround, I would not go for `track_features`.
I favor `conda-forge/label/debug::python` (or the `conda-forge/label/debug::debug` dependency if people have good reasons for it).

nikhilweee:
@mbargull Can you point me in the right direction on how to push to a separate label? I guess I'll need to pass on the `--label` argument to `conda mambabuild` but I can't edit `build_steps.sh` since that's overwritten by `conda-smithy`.

nikhilweee:
Sorry I didn't intend to mass-remove reviewers and now I don't see a way to add them back. Apologies.

nikhilweee:
I think at this point all except windows debug builds should be passing. I have no idea how to solve that and neither am I able to fully understand https://github.com/conda-forge/python-feedstock/issues/73#issuecomment-605151779 so I guess I could really use some help.

jakirkham:
No worries. This seems to still be in progress IIUC. We can add reviewers back when needed

jakirkham:
> I think at this point all except windows debug builds should be passing.

From [the Python docs]( https://docs.python.org/3/library/sys.html#sys.abiflags ):

> `sys.`**`abiflags`**
>
>    On POSIX systems where Python was built with the standard configure script, this contains the ABI flags as specified by [PEP 3149](https://peps.python.org/pep-3149/).

So my guess is we have some test of `sys.abiflags` somewhere that needs to be skipped on Windows. That said, atm am not finding the test here (though could be overlooking something). [These lines]( https://github.com/conda-forge/python-feedstock/blob/018b46521412f51cf732a1e4e08e999cb160f6bd/recipe/run_test.py#L94-L95 ) appear to be run right before the error (if that provides any clues)

nikhilweee:
@jakirkham Yes you're right, I added those tests [as part of this PR](https://github.com/conda-forge/python-feedstock/pull/597/commits/da448ec3e659562d2d908304efc7993048080d7a). In fact testing `sys.abiflags` and `sys. gettotalrefcount` are the only tests that are unique to the debug build. I'll change them to not run on Windows.

jakirkham:
> Yes you're right, I added those tests [as part of this PR](https://github.com/conda-forge/python-feedstock/pull/597/commits/da448ec3e659562d2d908304efc7993048080d7a).

🤦‍♂️ Yeah that makes sense. Didn't actually look here 😅 Probably time to call it a night 😴

> ...but then we'll have no way to test whether Windows debug builds are correct in any shape or form.

Tried searching for an alternative, but was unsuccessful (though also don't have a Windows machine to play with). There probably is an alternative, but it may require more searching or trying on a Windows machine. One can inspect flags Python was built with for example.

recipe/conda_build_config.yaml
```diff
@@ -6,3 +6,6 @@ numpy:
   - 1.16
 MACOSX_SDK_VERSION:            # [osx and x86_64]
   - 11.0                       # [osx and x86_64]
+build_type:
+  - release
+  - debug
```
mbargull, line 11: > @mbargull Can you point me in the right direction on how to push to a separate label? I guess I'll need to pass on the `--label` argument to `conda mambabuild` but I can't edit `build_steps.sh` since that's overwritten by `conda-smithy`.

You can add
```suggestion
build_type:
  - release
  - debug
+channel_targets:
+  - conda-forge main
+  - conda-forge debug
zip_keys:
  -
    - build_type
    - channel_targets
```
and rerendering should take care of everything ;).

mbargull:

recipe/conda_build_config.yaml
nikhilweee: Why do we use a nested list for zip keys? Can't I just do the following?
```yaml
zip_keys:
  - build_type
  - channel_targets
```

recipe/conda_build_config.yaml
```diff
@@ -6,3 +6,12 @@ numpy:
...
+channel_targets:
+  - conda-forge main
+  - conda-forge debug
+zip_keys:
+  - build_type
+  - channel_targets
```
mbargull, line 17: This has to be a list of lists.
```suggestion
zip_keys:
  - - build_type
    - channel_targets
```

recipe/conda_build_config.yaml
mbargull: (Didn't see your comment before.)
The `zip_keys` can have multiple entries (hence a list) and each entry is a list of keys to zip. You can see in the generated `.ci_support/*.yaml` the `c_compiler_version`/`cxx_compiler_version` as another entry being zipped.

recipe/conda_build_config.yaml
```diff
@@ -13,5 +13,6 @@ channel_targets:
   - conda-forge main
   - conda-forge debug
 zip_keys:
-  - - build_type
+  - 
+    - build_type
```
mbargull, line 17: BTW: No typo -- either is fine. But you can leave it as is now ;).

mbargull:

recipe/conda_build_config.yaml
nikhilweee: Oh I see, sorry I referred to the [docs](https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#coupling-keys) and that is what it suggested. I'm still curious to know why we use a nested list in `zip_keys` when the docs suggest either is fine?

nikhilweee:
@jakirkham @mbargull I think at this point all builds are passing. Is this ready to be merged?

mbargull:
@nikhilweee, thanks for working on this! Let's give others time to review before we proceed.

jakirkham:
Do we want this in a different label or do we just want some variant to select debug (like we do with PyPy, MPI, BLAS, etc.)?

recipe/meta.yaml
```diff
@@ -269,8 +269,11 @@ outputs:
       activate_in_script: true
       ignore_run_exports:
         - python_abi
+      script_env:
+        - PY_INTERP_DEBUG={{ py_interp_debug }}
       string: h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}{{ linkage_nature }}{{ debug }}_cpython  # ["conda-forge" in (channel_targets or "")]
       string: h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}{{ linkage_nature }}{{ debug }}  # ["conda-forge" not in (channel_targets or "")]
+      skip: true  # [win and build_type == "debug"]
```
mbargull, line 276: Same as above.

mbargull (dismissed):

mbargull:
> Do we want this in a different label or do we just want some variant to select debug (like we do with PyPy, MPI, BLAS, etc.)?

I'm -1 to a variant in the main channel. See https://github.com/conda-forge/python-feedstock/pull/597#pullrequestreview-1180378693 , specifically:
> The main cases for `track_features` are when we want certain variants/packages not installed by default but still have the possibility so seamlessly introduce them even implicitly (e.g. via dependencies).
> Since we shouldn't have the need for that implicit behavior and `track_features` is a non-perfect workaround, I would not go for `track_features`.

For one, it is usually undesirable to install a debug build unless you really want one. But also bare in mind that this isn't doesn't create a simple drop-in variant: the shared objects have another ABI tag, i.e., `311d` instead of just `311` (the suffix for extensions would also be `cpython-311d-x86_64-linux-gnu.so` and similar).

mbargull:
> > But also bare in mind that this isn't doesn't create a simple drop-in variant: the shared objects have another ABI tag, i.e., 311d instead of just 311 (the suffix for extensions would also be cpython-311d-x86_64-linux-gnu.so and similar).
> 
> That doesn't mean that extensions compiled with a release build cannot be loaded.

Right, I only meant the extensions that have been built with the debug version.

jakirkham:
Our knowledge of how to use features, particularly with Python (PyPy being a great example), has gotten a lot better over the years. Think we could go that route safely if we wanted to

nikhilweee:

mbargull:
> Our knowledge of how to use features, particularly with Python (PyPy being a great example), has gotten a lot better over the years. Think we could go that route safely if we wanted to

It's not about knowledge but (solver) logic. I'm quite familiar how the solver works; my deep dive into it was over 4 years ago and it still works the same.
Conda's strategy is to offer a solution if possible even if it isn't a good one. It'll take escape hatches if there are any.
We only use `track_features` because no one proposed/implemented a sound strategy to solve the use cases we (ab-)use `track_features` now (the original design was different from how we use them now).
We don't have to reiterate all of this; but here are a simple demonstration of some of its pitfalls:
```sh
(base) conda create -y --name=test >/dev/null 2>&1
(base) conda activate test
(test) # Pretend we'd have some package with a constraint that
(test) #   - prevents installation of python[version=>=3.9.10,build=*_cpython]
(test) #   - allows installations of python[version=>=3.9.10,build=*_pypy]
(test) conda config --env --add pinned_packages 'ld_impl_linux-64<2.36'
(test) # In the following, *_pypy packages are never requested per CLI nor history specs.
(test) # ... we still end up and get stuck ("features" linger per solver logic!) with them anyway...
(test) conda install -qd python=3.9.10 2>&1 | grep 'python-3'
  python             conda-forge/linux-64::python-3.9.10-0_73_pypy None
(test) mamba install -qy python=3.9.10 2>&1 | grep 'python *3'
  + python              3.9.10  0_73_pypy      conda-forge/linux-64     Cached
(test) # Lift the constraint -- *_cpython packages are still not installed by default
(test) # (as per design from the track_features mechanism!)
(test) conda config --env --remove-key pinned_packages
(test) conda install -qd python=3.9.12 2>&1 | grep 'python *3'
  python                                   3.9.10-0_73_pypy --> 3.9.12-0_73_pypy None
(test) mamba install -qd python=3.9.12 2>&1 | grep 'python *3'
  - python    3.9.10  0_73_pypy   conda-forge                    
  + python    3.9.12  0_73_pypy   conda-forge/linux-64     Cached
(test) conda install -qd python=3.9.12=\*cpython 2>&1 | grep 'python *3'
  python                                   3.9.10-0_73_pypy --> 3.9.12-h2660328_1_cpython None
(test) mamba install -qd python=3.9.12=\*cpython 2>&1 | grep 'python-3'
  - package pypy3.9-7.3.8-h986b250_2 has constraint python 3.9.* *_73_pypy conflicting with python-3.9.12-h9a8a25e_0_cpython
```
We do use `track_features` to have an easy way for users to install, e.g., PyPy.
But it is still a brittle construct -- most of the time it works, but it can also fail easily (as demonstrated above).
As stated before, with debug builds we (IMO) don't have the need to rely on implicit, sometimes working behavior, i.e., provide an "easy way for users to install" them, but can use a proper, logically more sound way (i.e., separate channels).
With `track_features` there is always a (albeit small) possibility for regular users to end up with the debug builds; with separate channels there is none. Regular users don't need the debug builds; hence "none" is a better choice here.

nikhilweee:
@mbargull How long does the review process generally take?

nikhilweee:
It's also strange that we still have windows debug builds instead of release builds even when `recipe.yaml` explicitly skips windows debug builds.

recipe/build_base.sh
```diff
@@ -508,7 +508,8 @@ fi
 rm -rf ${PREFIX}/lib/python${VER}/distutils/command/*.exe
 
 python -c "import compileall,os;compileall.compile_dir(os.environ['PREFIX'])"
-rm ${PREFIX}/lib/libpython${VER}.a
+rm ${PREFIX}/lib/libpython${VERABI}.a
+
```
nikhilweee, line 512: I wonder if it also makes sense to
```
ln -s ${PREFIX}/lib/libpython${VER}d.so ${PREFIX}/lib/libpython${VER}.so
```

I suggest this because gdb complaints about the following:
```
gdb: error while loading shared libraries: libpython3.10.so.1.0: cannot open shared object file: No such file or directory
```

recipe/build_base.sh
```diff
@@ -508,7 +508,8 @@ fi
...
 if [[ "$target_platform" == linux-* ]]; then
```
nikhilweee, line 513: The `bin` directory has two executables of the same size. How about making one a symlink of the other?
```
if [[ ${PY_INTERP_DEBUG} == yes ]]; then
  rm ${PREFIX}/bin/python{VER}
  ln -s ${PREFIX}/bin/python{VERABI} ${PREFIX}/bin/python{VER}
fi
```

nikhilweee:
Im also curious to know why we ship `__pycache__` and `*.pyc`? This may be a really dumb question but aren't those files regenerated automatically?

recipe/build_base.sh
```diff
@@ -97,6 +97,12 @@ if [[ ${_OPTIMIZED} = yes ]]; then
   CXXFLAGS=$(echo "${CXXFLAGS}" | sed "s/-O2/-O3/g")
 fi
 
+if [[ ${PY_INTERP_DEBUG} == yes ]]; then
+  CPPFLAGS=$(echo "${CPPFLAGS}" | sed "s/-O2/-O0/g")
+  CFLAGS=$(echo "${CFLAGS}" | sed "s/-O2/-O0/g")
+  CXXFLAGS=$(echo "${CXXFLAGS}" | sed "s/-O2/-O0/g")
+fi
+
```
nikhilweee, line 105: Does it make sense to disable compile time optimizations for debug builds?

chrisburr:
> Im also curious to know why we ship __pycache__ and *.pyc? This may be a really dumb question but aren't those files regenerated automatically? In my analysis we can further save some space by removing those files.

This is how it's normally done by package managers (for example [see the fedora's Python 3 file list](https://fedora.pkgs.org/37/fedora-x86_64/python3-libs-3.11.0-1.fc37.x86_64.rpm.html)). There are a few reasons (and I'm probably forgetting others):

* don't want to be executing Python at install time to regenerate them (you might not even be able to if you're installing from a different platform)
* generating on demand might not be possible if the environment isn't writable
* generating on demand results in files being left behind when you uninstall packages
* adding Python-specific workarounds to the package manager becomes problematic when Python changes as you suddenly have old versions of conda doing the wrong thing

nikhilweee:
@chrisburr Those are really great points! Thank you for the explanation.

nikhilweee:

nikhilweee:
@conda-forge-admin, please ping team

It's been a while since this PR is open and It would be great to get an opinion from the team before we move forward. If this is left unaddressed I'm afraid this PR will fall behind other commits and I might not be able to keep up with it.

nikhilweee:
I also wanted to know what's the best way to incorporate upstream changes? Right now I did a `rebase` but I'm not sure if that's the recommended way.

mbargull:
> I also wanted to know what's the best way to incorporate upstream changes? Right now I did a `rebase` but I'm not sure if that's the recommended way.

There are multiple ways but no recommended way.
Personally, for our feedstocks, esp. the more involved and important ones like this one, I find it easiest for the reviewers to have a somewhat clear change set, i.e., few unrelated changes and separable commits. That can be achieved with rebases/cherry-picking/squashes, if needed -- the downside is that needed force-pushes for that can make previous CI runs and their logs less visible; hence one should make deliberate/sparingly use of those.

In this case, we have the history of different attempts and some combination of merged and rebased commit which leaves us with a long list of commits that takes a bit of time to comprehend. Hence, I'd probably reorder+squash related commits and cherry-pick your changes on top of our `main` branch to make this now more comprehensible.

recipe/build_base.sh
mbargull: Yes. There are cases in which an optimized debug build does make sense, but an `-O0` build is probably what most would expect.

recipe/build_base.sh
mbargull: Yes, I checked the artifact and the files are indeed identical, so this is probably fine.

recipe/meta.yaml
```diff
@@ -84,6 +84,7 @@ source:
 
 build:
   number: {{ build_number }}
+  skip: true  # [win and build_type == "debug"]
```
mbargull, line 87: I haven't yet looked into what caused the rendering issue with `conda-smithy`, but maybe we can try
```suggestion
  skip: true  # [win and build_type != "release"]
```
instead.

mbargull: Oh, I just noticed that the last rerendering messed up the non-Windows builds, too. Need to investigate...

nikhilweee: Yeah this PR has become a mess. I need to clean it up a bit.

mbargull:

nikhilweee:

nikhilweee:
Okay, I'm not very happy about force pushing, but the history should be much cleaner now.

nikhilweee:

recipe/tests/cmake/CMakeLists.txt
```diff
@@ -1,11 +1,9 @@
 # https://martinopilia.com/posts/2018/09/15/building-python-extension.html
+# FindPythonInterp and FindPythonLibs is deprecated since cmake 3.12
```
isuruf, line 2: That doesn't mean it's not used by projects. Please keep the test as it is and add another if you want to.

recipe/run_test.py
```diff
@@ -86,17 +86,20 @@
     import syslog
     import termios
 
+    if os.getenv('PY_INTERP_DEBUG') == 'yes':
+        assert 'd' in sys.abiflags
+        assert 'gettotalrefcount' in dir(sys)
```
isuruf, line 91: Can you check this for windows too?

recipe/meta.yaml
```diff
@@ -26,14 +26,14 @@
...
   {% set linkage_nature_env = 'static' %}
 {% endif %}
-{% set dbg_abi = "" %}
-{% set debug = os.environ.get('PY_INTERP_DEBUG', '') %}
-{% if debug != '' and debug != 'no' %}
+{% if build_type == "debug" %}
   {% set py_interp_debug = "yes" %}
-  {% set debug = "_dbg" %}
+  {% set debug = "_debug" %}
   {% set dbg_abi = "d" %}  # [win]
```
isuruf, line 32: This is not true anymore. You can do `dbg_abi = ""` for all platforms.

isuruf (changes requested):
Please add a run_exports that would require a debug python in downstream pacakges for the debug package.

recipe/run_test.py
nikhilweee: Please see https://github.com/conda-forge/python-feedstock/pull/597#issuecomment-1314960208. `assert 'd' in sys.abiflags` only works for POSIX systems. This is also why debug builds are currently skipped for windows. Not sure if the other test (`'gettotalrefcount' in dir(sys)`) works on windows though.

isuruf: `sys.abiflags` does not change in windows because the headers are not configured unlike in unix. However `sys.gettotalrefcount` is a good test.

recipe/tests/cmake/CMakeLists.txt
nikhilweee: Sure, I'll add them back.

nikhilweee: @isuruf could you maybe have a look at the CI logs and help me understand why the builds are failing with `FindPythonInterp` and `FindPythonLibs`? That's why I switched to `FindPython` in the first place.

jakirkham:
Could you please explain why that `run_export` should be added? Asking as it is my understanding this should be ABI compatibility on UNIX (all that is built now). Also this has been segregated into a different label (where other packages that would build against likely would need to go). That said, not sure that we are going to build packages dependent on the debug build of this package (since we can reuse their existing builds)

isuruf:
See https://github.com/conda-forge/python-feedstock/pull/597#issuecomment-1315925092 and the two replies afterwards.

jakirkham:
Not seeing the relationship. Am hearing Marcel wants to stick with a label

isuruf:
>  the suffix for extensions would also be cpython-311d-x86_64-linux-gnu.so and similar

Extensions built with debug cannot be imported in release

jakirkham:
Right are we planning to build things with the debug `python` build? My understanding was no

recipe/meta.yaml
```diff
@@ -128,6 +128,9 @@ outputs:
       string: h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}{{ linkage_nature }}{{ debug }}  # ["conda-forge" not in (channel_targets or "")]
 {% if 'conda-forge' in channel_targets %}
       run_exports:
+{% if py_interp_debug == "yes" %}
+        - python=*=*debug*
+{% endif %}
```
nikhilweee, line 133: @isuruf Is this what you were suggesting or is there a better way?

nikhilweee: Turns out this is invalid syntax so I guess I need help here.

nikhilweee:
@isuruf Please have a look at the unresolved conversations. I need help with figuring out:
1. why CMake tests fail on debug builds (see [relevant CI logs](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=616844&view=results)) and 
2. how to add a `run_exports` for the debug build.

recipe/build_base.sh
```diff
@@ -520,7 +520,7 @@ if [[ ${PY_INTERP_DEBUG} == yes ]]; then
   rm ${PREFIX}/bin/python${VER}
   ln -s ${PREFIX}/bin/python${VERABI} ${PREFIX}/bin/python${VER}
   ln -s ${PREFIX}/lib/libpython${VERABI}.so ${PREFIX}/lib/libpython${VER}.so
-  ln -s ${PREFIX}/lib/libpython${VERABI}.so.1.0 ${PREFIX}/lib/libpython${VER}.so.1.0
+  ln -s ${PREFIX}/include/python${VERABI} ${PREFIX}/include/python${VER}
```
nikhilweee, line 523: @isuruf Thanks! How were you able to figure this out? I was having a really hard time

recipe/meta.yaml
nikhilweee: Solved in future commit.

recipe/build_base.sh
isuruf: Error message said
`Could NOT find PythonLibs (missing: PYTHON_INCLUDE_DIRS) (Required is at least version "3.11.0")`

recipe/tests/cmake/CMakeLists.txt
nikhilweee: Resolved in future commit

recipe/build_base.sh
nikhilweee: I see, I guess these kinds of errors become easy to address with experience, something that I lack right now.

isuruf:
> Right are we planning to build things with the debug python build? My understanding was no

No, but it is good to be correct.

nikhilweee:

nikhilweee:
@conda-forge-admin, please ping team

Hi team, this PR has had no activity for quite sometime. I wonder what are the next steps?

hmaarrfk:
Did this break cross python? We seem to be having trouble with python 3.11 + osx cross python that go away if we pin to _0_cpython

xref: https://github.com/conda-forge/pytorch-cpu-feedstock/pull/158

hmaarrfk:
cc: @ngam

isuruf:
What's the error?

isuruf:
Should be fixed now. There was an error in rerendering that messed up the labels in anaconda.org.

hmaarrfk:
Thank you!

PR #609, opened 2022-12-08
Update to 3.7.16

moreshiny:
Checklist


moreshiny:

moreshiny:

moreshiny:
Would anyone have chance to look this over?

hmaarrfk:
Generally speaking, we've stopped supporting python 3.7. So renaming all the patches is somewhat disruptive for the review process

hmaarrfk:
There seems to be a few patches that you've made a marked improvement on. It would be good if in the pull request, you left everything you could untouched. i think you can specify a starting number for the git format patch command. that should alleviate alot of the changes.

moreshiny:
@hmaarrfk Yeah the reordering and cleanup changes mostly came from me fighting with the patches but the most minimal set of changes is actually quite small and related to zlib/bzip2/openssl version changes.

I've reduced it to that. This means that the chunk ranges are not correct but the patches apply anyway. Let me know if I should bring those and/or the patch index changes back in.

recipe/patches/0016-Unvendor-openssl.patch
```diff
@@ -75,24 +75,24 @@ index b43c674..16d00fc 100644
...
      <sqlite3Dir>$(ExternalsDir)sqlite-3.31.1.0\</sqlite3Dir>
-     <bz2Dir>$(ExternalsDir)bzip2-1.0.6\</bz2Dir>
+     <bz2Dir>$(ExternalsDir)bzip2-1.0.8\</bz2Dir>
      <lzmaDir>$(ExternalsDir)xz-5.2.2\</lzmaDir>
--    <opensslDir>$(ExternalsDir)openssl-1.1.1g\</opensslDir>
--    <opensslOutDir>$(ExternalsDir)openssl-bin-1.1.1g\$(ArchName)\</opensslOutDir>
+-    <opensslDir>$(ExternalsDir)openssl-1.1.1n\</opensslDir>
+-    <opensslOutDir>$(ExternalsDir)openssl-bin-1.1.1n\$(ArchName)\</opensslOutDir>
 -    <opensslIncludeDir>$(opensslOutDir)include</opensslIncludeDir>
 +    <opensslDir>$(OPENSSL_DIR)\</opensslDir>
```
hmaarrfk, line 91: can you explain in a few words why this change is necessary now? it uses an environment variable which is new.

hmaarrfk:
@conda-forge/python I want to bring to your attention that this PR would introduce the breaking str->int conversion change

I know we decided 
* https://conda-forge.org/docs/user/announcements.html#releasing-python-3-8-14-3-9-14-and-3-10-7
* https://github.com/conda-forge/python-feedstock/pull/579

to release these on the supported versions of python, but i'm not sure what we want to do here, given that we "stopped supporting" python 3.7.

I'm ok with releasing this, but I just wanted to bring it up.

https://docs.python.org/3.7/whatsnew/changelog.html#python-3-7-14-final

recipe/patches/0016-Unvendor-openssl.patch
moreshiny: I have not changed line 91 with the OPENSSL_DIR variable, this has been in the patch file for a long time. It just looks that way because it's a diff of a patch file which Github doesn't display that well.

hmaarrfk: understood. perfect. Thank you for pointing that out.

It seems that python changed the version of openssl they built against.

Our patch also previously replaced it with an environment variable.

hmaarrfk (approved):

isuruf:
Please use https://github.com/conda-forge/python-feedstock/blob/main/recipe/patches/README.md when updating the patches here.

hmaarrfk:
> Please use https://github.com/conda-forge/python-feedstock/blob/main/recipe/patches/README.md when updating the patches here.

Sorry for the noise on reformatting the patches. I defer to Isuru's expertise here.

moreshiny:
> > Please use https://github.com/conda-forge/python-feedstock/blob/main/recipe/patches/README.md when updating the patches here.
> 
> Sorry for the noise on reformatting the patches. I defer to Isuru's expertise here.

No problem at all.

@isuruf Just to be clear, I would:
- include all changes to all patches from that procedure in the Readme
- renumber the patches to start at 0001
- use the procedure with `git format-patch --no-signature` https://github.com/conda-forge/python-feedstock/blob/3.10/recipe/patches/README.md

Let me know if that's ok for you then I will regenerate all patches on this and the other version PRs (and add/update the patch Readme as needed).

jakirkham:
@conda-forge-admin, please re-render

PR #604, opened 2022-11-21
updated v3.8.15

mbargull:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #618, opened 2023-01-26
Rebuild 3.8.15

h-vetinari:
After merging #604, one of the PPC builds failed, and cannot now be rebuilt due to breaking changes in the setuptools `downstreams:` test.

@jakirkham [asked](https://github.com/conda-forge/python-feedstock/pull/604#issuecomment-1404717531) me to send a PR to fix it.


recipe/meta.yaml
```diff
@@ -187,9 +187,9 @@ outputs:
 {% endif %}
     test:
       downstreams:
-        # Temporarily disable dowmstream tests on aarch64 as the solver causes timeouts on drone
-        - cython  # [not aarch64]
-        - setuptools  # [not aarch64]
+        # Temporarily disable downstream test for setuptools due to breaking changes
```
jakirkham, line 190: Let's keep the old comment and just add a new one. Since at some point, we presumably want to add these tests back

```suggestion
        # Temporarily disable dowmstream tests on aarch64 as the solver causes timeouts on drone
        # Also temporarily skipping downstream test for setuptools due to breaking changes
```

recipe/meta.yaml
```diff
@@ -187,9 +187,9 @@ outputs:
...
+        - cython
```
jakirkham, line 191: Guessing we still want the `not aarch64` selector due to the previous issue

```suggestion
        - cython  # [not aarch64]
```

recipe/meta.yaml
```diff
@@ -187,9 +187,9 @@ outputs:
...
+        # - setuptools
```
jakirkham, line 192: Would keep the selector and just comment this line

```suggestion
        # - setuptools  # [not aarch64]
```

jakirkham (commented):
Thanks Axel! 🙏

Had a few suggestions below as I think we want to keep the ARM skips in the future (once we can revert this). This would also be consistent with [how this was handled in `main`]( https://github.com/conda-forge/python-feedstock/blob/21038d707640da7b5da152a5239e199c720f0574/recipe/meta.yaml#L191-L195 ) for the Python 3.11 migration

recipe/meta.yaml
h-vetinari: Well, the comment was referencing drone, which we're not using anymore. Might as well try if it still times out. :)

recipe/meta.yaml
h-vetinari: I commented out the whole line because AFAICT, the setuptools-breakage is not confined to aarch/ppc

h-vetinari:
> This would also be consistent with [how this was handled in `main`](https://github.com/conda-forge/python-feedstock/blob/21038d707640da7b5da152a5239e199c720f0574/recipe/meta.yaml#L191-L195) for the Python 3.11 migration

I guess I can comment out the downstream tests completely, but since I was touching it, I tried to go for not removing tests if it isn't necessary. If it fails, I'll skip them.

recipe/meta.yaml
jakirkham: Sure if you'd like to spend time trying/fixing it, that would of course be welcome :)

Though completely understand if you would rather stick to just getting the missing builds out

recipe/meta.yaml
jakirkham: Currently yes. Though if in the future the `setuptools` issue is fixed, but the `aarch64` issue persists, it may be helpful for the next person who comes here to back things out to the last known working state

recipe/meta.yaml
h-vetinari: It passed on travis, so that should be good enough to keep this bit. ;-)

recipe/meta.yaml
h-vetinari: I don't expect the drone timeouts to be relevant anymore, so resolving this.

jakirkham:
Can we please bump the `build/number` as well? Changing `test` effects the packages. Especially when passing to `conda build --test <pkg>`

PR #641, opened 2023-09-19
Add 3.10 and 3.11 to abi migrations

xhochy:
I just created 3.11 and also quoted the branches to avoid YAML parsing issues.

### DIFF
--- a/conda-forge.yml
+++ b/conda-forge.yml
@@ -4,7 +4,6 @@ provider: {linux_aarch64: default, linux_ppc64le: native}
 test_on_native_only: true
 bot:
   abi_migration_branches:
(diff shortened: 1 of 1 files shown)


conda-forge.yml
```diff
@@ -4,9 +4,11 @@ provider: {linux_aarch64: default, linux_ppc64le: native}
 test_on_native_only: true
 bot:
   abi_migration_branches:
-  - 3.7
-  - 3.8
-  - 3.9
+  - "3.7"
```
h-vetinari, line 7: We can remove 3.7, it's been dropped in conda-forge since last October (and EOL upstream since this July)

PR #648, opened 2023-09-24
Add runtime _python_rc dependency for rc builds and publish to main

0xbe7a:
- publish to main channel
- issue has been closed
- MNT: Re-rendered with conda-build 3.25.0, conda-smithy 3.26.2, and conda-forge-pinning 2023.09.23.18.20.25

Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

See https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/280

CC @isuruf @xhochy

PR #636, opened 2023-09-01
Python 3.12.0rc2

xhochy:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #651, opened 2023-10-02
Python 3.12.0

h-vetinari:
Essentially all the work was done in #636, plus a little bit in #642 & #648. Due to the switch introduced in #648 
https://github.com/conda-forge/python-feedstock/blob/7614a9b8443e983309ae97805abf86eb67696eb3/recipe/meta.yaml#L182-L184

we don't have to update the channels for this PR.

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -8,11 +8,11 @@ jobs:
     vmImage: ubuntu-latest
   strategy:
     matrix:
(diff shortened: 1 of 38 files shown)


h-vetinari:
@conda-forge/core, Python 3.12 was released, and the update PR is nice and painless. The question is just (IMO) when we want things to be available.

hmaarrfk (commented):
great work!

isuruf:
@conda-forge-admin, rerender

PR #656, opened 2023-12-10
Update to Python 3.12.1

ismail:
Refresh 0005-Unvendor-openssl.patch to apply

Checklist

### DIFF
--- a/.ci_support/osx_64_build_typereleasechannel_targetsconda-forge_python_debug.yaml
+++ b/.ci_support/osx_64_build_typereleasechannel_targetsconda-forge_python_debug.yaml
@@ -1,33 +1,29 @@
-BUILD:
-- aarch64-conda_cos7-linux-gnu
+MACOSX_DEPLOYMENT_TARGET:
+- '10.9'
+MACOSX_SDK_VERSION:
+- '11.0'
 build_type:
-- debug
+- release
 bzip2:
 - '1'
 c_compiler:
-- gcc
+- clang
 c_compiler_version:
-- '12'
-cdt_arch:
-- aarch64
-cdt_name:
-- cos7
+- '16'
 channel_sources:
 - conda-forge
 channel_targets:
 - conda-forge python_debug
 cxx_compiler:
-- gxx
+- clangxx
 cxx_compiler_version:
-- '12'
-docker_image:
-- quay.io/condaforge/linux-anvil-aarch64
+- '16'
 expat:
 - '2'
 libffi:
 - '3.4'
-libuuid:
-- '2'
+macos_machine:
+- x86_64-apple-darwin13.4.0
 ncurses:
 - '6'
 openssl:
@@ -43,7 +39,7 @@ readline:
 sqlite:
 - '3'
 target_platform:
-- linux-aarch64
+- osx-64
 tk:
 - '8.6'
 xz:
--- a/recipe/patches/0022-Allow-linking-ncursesw-panelw-on-macOS-if-available.patch
+++ b/recipe/patches/0022-Allow-linking-ncursesw-panelw-on-macOS-if-available.patch
@@ -1,0 +1,85 @@
+From 0d6fe6bc4e993b0e9c51edf78a5727e779c9b6ce Mon Sep 17 00:00:00 2001
+From: Marcel Bargull <marcel.bargull@udo.edu>
+Date: Thu, 14 Dec 2023 22:16:33 +0100
+Subject: [PATCH 22/22] Allow linking ncursesw/panelw on macOS if available
+
+Signed-off-by: Marcel Bargull <marcel.bargull@udo.edu>
+---
+ configure    | 4 ----
+ configure.ac | 6 ------
+ 2 files changed, 10 deletions(-)
+
+diff --git a/configure b/configure
+index 4155e2fe25..22eb0cd52f 100755
+--- a/configure
++++ b/configure
+@@ -25688,7 +25688,6 @@ fi
+ if test "x$ac_cv_header_ncurses_h" = xyes
+ then :
+ 
+-  if test "$ac_sys_system" != "Darwin"; then
+ 
+ pkg_failed=no
+ { printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for ncursesw" >&5
+@@ -25884,7 +25883,6 @@ printf "%s\n" "yes" >&6; }
+       have_curses=ncursesw
+ 
+ fi
+-  fi
+ 
+   if test "x$have_curses" = xno
+ then :
+@@ -26120,7 +26118,6 @@ if test "x$ac_cv_header_panel_h" = xyes
+ then :
+ 
+ 
+-  if test "$ac_sys_system" != "Darwin"; then
+         if test "x$have_curses" = xncursesw
+ then :
+ 
+@@ -26315,7 +26312,6 @@ printf "%s\n" "yes" >&6; }
+ fi
+ 
+ fi
+-  fi
+ 
+   if test "x$have_curses" = xncurses
+ then :
+diff --git a/configure.ac b/configure.ac
+index 1bf0fa8ad0..6ff8b4a2dc 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -6405,8 +6405,6 @@ AH_TEMPLATE([HAVE_NCURSESW], [Define to 1 if you have the `ncursesw' library.])
+ AC_CHECK_HEADERS([curses.h ncurses.h])
+ 
+ AS_VAR_IF([ac_cv_header_ncurses_h], [yes], [
+-  if test "$ac_sys_system" != "Darwin"; then
+-    dnl On macOS, there is no separate /usr/lib/libncursesw nor libpanelw.
+     PKG_CHECK_MODULES([CURSES], [ncursesw], [
+       AC_DEFINE([HAVE_NCURSESW], [1])
+       have_curses=ncursesw
+@@ -6420,7 +6418,6 @@ AS_VAR_IF([ac_cv_header_ncurses_h], [yes], [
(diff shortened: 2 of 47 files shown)


ismail:

ismail:

ismail:
macOS builds fail with an ncurses related error

```
[...]/lib-dynload/_curses.cpython-312-darwin.so, 2): Symbol not found: _extended_color_content
``

**Comment by ismail** [comment]
So, Python has this check to see if it can use `_extended_color_content`:

```
#if NCURSES_EXT_FUNCS+0 >= 20170401 && NCURSES_EXT_COLORS+0 >= 20170401
#define _NCURSES_EXTENDED_COLOR_FUNCS   1
#else
#define _NCURSES_EXTENDED_COLOR_FUNCS   0
#endif

#if _NCURSES_EXTENDED_COLOR_FUNCS
#define _CURSES_COLOR_VAL_TYPE          int
#define _CURSES_COLOR_NUM_TYPE          int
#define _CURSES_INIT_COLOR_FUNC         init_extended_color
#define _CURSES_INIT_PAIR_FUNC          init_extended_pair
#define _COLOR_CONTENT_FUNC             extended_color_content
#define _CURSES_PAIR_CONTENT_FUNC       extended_pair_content
#else
#define _CURSES_COLOR_VAL_TYPE          short
#define _CURSES_COLOR_NUM_TYPE          short
#define _CURSES_INIT_COLOR_FUNC         init_color
#define _CURSES_INIT_PAIR_FUNC          init_pair
#define _COLOR_CONTENT_FUNC             color_content
#define _CURSES_PAIR_CONTENT_FUNC       pair_content
#endif  /* _NCURSES_EXTENDED_COLOR_FUNCS */
```

and if I check the Python 3.12 binary from Conda:

```
❯ nm -u ~/envs/stats/.pixi/env/lib/python3.12/lib-dynload/_curses.cpython-312-darwin.so | grep color
_can_change_color
_color_content
_has_colors
_init_color
_start_color
_use_default_colors
```

so this check failed and the non-extended version of the functions were used. My guess would be `ncurses` package was updated in Conda, and it's not working properly for macOS x86-64.

**Comment by ismail** [comment]
https://invisible-island.net/ncurses/INSTALL.html says

```
    --enable-ext-colors
        Extend the cchar_t structure to allow more than 16 colors to be
        encoded.  This applies only to the wide-character (--enable-widec)
        configuration.

        NOTE: using this option will make libraries which are not binary-
        compatible with libncursesw 5.4.  None of the interfaces change, but
        applications which have an array of cchar_t's must be recompiled.
```

so in the https://github.com/conda-forge/ncurses-feedstock/blob/main/recipe/build.sh we should change

```
WIDEC_OPT="--enable-widec"
```

to

```
WIDEC_OPT="--enable-widec --enable-ext-colors"
```

I think this is getting enabled for arm64, but not for x86_64.

**Comment by ismail** [comment]
OK, not so fast. Looking at the ncurses6 package from conda-forge

```
❯ file ./.pixi/env/lib/libncurses.6.dylib
./.pixi/env/lib/libncurses.6.dylib: Mach-O 64-bit arm64 dynamically linked shared library, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL>

❯ nm ./.pixi/env/lib/libncurses.6.dylib | grep init_color
00000000000047b4 T __nc_init_color
00000000000049ac T _init_color
00000000000049a8 T _init_color_sp
```

This feature is not enabled on arm64 either, Python must be misdetecting it.

But if I look at the arm64 binary for Python 3.12.1:

```
❯ nm -u ./osx-arm64/lib/python3.12/lib-dynload/_curses.cpython-312-darwin.so | grep color
_can_change_color
_extended_color_content
_has_colors
_init_extended_color
_start_color
_use_default_colors
```

so it's using extended colors and the test is not failing. Very confusing.

**Comment by ismail** [comment]
Apparently until https://github.com/python/cpython/commit/d34650e7409b6a7cc64d7a118e4500c2f3cdbcf3 Python3 was not correctly detecting `ncursesw` on macOS, so this explains why this was not seen before.

**Comment by ismail** [comment]
Someone else, please feel free to pick this up.

I think building non-wide curses on macOS doesn't make sense, but it'd be too big of a hammer. We just need to make sure Python links to `ncursesw` on macOS.

File: recipe/patches/0022-Allow-linking-ncursesw-panelw-on-macOS-if-available.patch
Comment (mbargull) [inline recipe/patches/0022-Allow-linking-ncursesw-panelw-on-macOS-if-available.patch:1]: There are multiple ways we can go about this:
1. Use this patch.
2. Upstream this patch.
3. Change our `ncurses` builds to make wide character support available from the non-`*w` names libraries.

re. 1.:
It should restore pre-`3.12` linking behavior (look for `ncursesw`/`panelw` first and prefer those if found).
The added test command failed for `3.12.0` as expected at https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=844414&view=logs&j=f23f984a-2b8a-59f7-ed33-e22fa8dd9908&t=1d351711-1535-5fad-3bab-ff535f39fb55&l=11742 and confirms that `3.12.1` has proper wide character support in the `curses` module (one can also see in the logs when `ncurses` or `ncursesw` are linked to).
But maintaining such patch here is not optimal, of course.

re. 2.:
There might be some preference to link to `ncurses`/`panel` on macOS if they have wide character support (since those from Apple/Homebrew/MacPorts builds do).
This would mean the order of preference would need to be reversed here (possibly some symbol checks added).
Note that for `<=3.11` there it would link to the `*w` libraries first, i.e., just like this patch does again; ref: https://github.com/python/cpython/commit/e925241d95d8095adf67f492042f97254ff82ec1#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L1106 .

re. 3.:
- Homebrew has symlinks from non-`*w` pointing to `*w` libraries:
    - https://github.com/Homebrew/homebrew-core/blob/24808e755b8005b39c59029507e9a43d4e5a8296/Formula/n/ncurses.rb#L65
- MacPorts have symlinks from `*w` poiting to non-`*w` libraries:
    - https://github.com/macports/macports-ports/blob/703e76e8f90d7fb0837b315cf414b4175b61fbbb/devel/ncurses/Portfile#L36
    - https://github.com/macports/macports-ports/blob/703e76e8f90d7fb0837b315cf414b4175b61fbbb/devel/ncurses/Portfile#L68
- Fedora- and Debian-based distributions seem to have separate non-`*w` libraries, i.e., without wide char support like `conda-forge::ncurses`.
- Arch Linux-based distributions use linker scripts for non`-*w` libraries pointing to the `*w` ones
    - https://gitlab.archlinux.org/archlinux/packaging/packages/ncurses/-/blob/3c2606603aa4a5a3b2d29e560a1bc14986153f49/PKGBUILD#L74

I don't know what (if any) downsides there are to not having the wide character supported merged in to the non-`*w` named libraries.
(^-- This should be copied into a discussion in the ncurses feedstock; noting it here for now so we have context to discuss on what to do with this patch.)

**Comment by mbargull** [comment]

File: .ci_support/osx_64_build_typereleasechannel_targetsconda-forge_python_debug.yaml [hunk 1]
```diff
@@ -3,7 +3,7 @@ MACOSX_DEPLOYMENT_TARGET:
 MACOSX_SDK_VERSION:
 - '11.0'
 build_type:
-- debug
+- release
```
mbargull, line 6: That ain't right...
Somehow the wrong values seem to have been zipped together.
I haven't yet checked whether I can reproduce this locally...
cc @conda-forge/core for awareness

mbargull:

.ci_support/osx_64_build_typereleasechannel_targetsconda-forge_python_debug.yaml
mbargull: Can't reproduce for now.
Welp, an adventure for another day, I guess...

mbargull:
@ismail, thanks for looking into this!
I've added a patch so that we link to `ncursesw` again and added a test so don't regress on this again.

> Apparently until [python/cpython@d34650e](https://github.com/python/cpython/commit/d34650e7409b6a7cc64d7a118e4500c2f3cdbcf3) Python3 was not correctly detecting `ncursesw` on macOS, so this explains why this was not seen before.

Yep, `conda-forge::python=3.12.0` definitely has wide character support broken on macOS; confirmed with https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=844414&view=logs&j=f23f984a-2b8a-59f7-ed33-e22fa8dd9908&t=1d351711-1535-5fad-3bab-ff535f39fb55&l=11742 .
The non-`3.12` builds should be fine, though, since it broke with https://github.com/python/cpython/commit/e925241d95d8095adf67f492042f97254ff82ec1 which is in `3.12.0a1` onward.

> which also *should* make arm64 test fail, but that's not happening.

Our macOS CI workers are currently x86-only such that we have to cross-compile the `osx-arm64` builds with no automated testing happening for them for now. Meaning, the tests didn't fail because they weren't run.

mbargull:
@conda-forge-admin, please restart ci

recipe/patches/0022-Allow-linking-ncursesw-panelw-on-macOS-if-available.patch
mbargull: re. 3.: Opened https://github.com/conda-forge/ncurses-feedstock/pull/76 -- that's still WIP (and should only be merged after careful review after the holidays).

mbargull:

mbargull:
My suggestion would be to merge this and revisit further possible actions from https://github.com/conda-forge/python-feedstock/pull/656#discussion_r1428491692 later.

@conda-forge/python, anyone else to review?

PR #664, opened 2023-12-23
Build against libxcrypt on Linux

mbargull:
refs:
- https://github.com/conda-forge/linux-sysroot-feedstock/issues/52

Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #668, opened 2023-12-23
[3.8] Build against libxcrypt on Linux

mbargull:
refs:
- https://github.com/conda-forge/python-feedstock/pull/664
- https://github.com/conda-forge/linux-sysroot-feedstock/issues/52

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -254,6 +254,8 @@ outputs:
         - test ! -f default.profraw   # [osx]
         # Test for segfault on osx-64 with libffi=3.4, see https://bugs.python.org/issue44556
         - python -c "from ctypes import CFUNCTYPE; CFUNCTYPE(None)(id)"
+        # Test for wide character supported via ncursesw
+        - TERM=xterm >/dev/null python -c "import curses; scr = curses.initscr(); curses.unget_wch('x'); assert 'x' ==  scr.get_wch()"  # [unix]
 
   - name: libpython-static
     script: build_static.sh  # [unix]
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -4,7 +4,7 @@
 {% set ver2 = '.'.join(version.split('.')[0:2]) %}
 {% set ver2nd = ''.join(version.split('.')[0:2]) %}
 {% set ver3nd = ''.join(version.split('.')[0:3]) %}
-{% set build_number = 0 %}
+{% set build_number = 1 %}
 
 # this makes the linter happy
 {% set channel_targets = channel_targets or 'conda-forge main' %}
@@ -179,6 +179,7 @@ outputs:
         - ld_impl_{{ target_platform }} >=2.36.1  # [linux]
         - libnsl  # [linux]
         - libuuid  # [linux]
+        - libxcrypt  # [linux]
       run:
         - ld_impl_{{ target_platform }} >=2.36.1  # [linux]
 {% if 'conda-forge' in channel_targets %}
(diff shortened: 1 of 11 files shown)


mbargull:
gh-664 was merged, so I'm putting in the version-branched PRs as well.

recipe/meta.yaml
```diff
@@ -179,6 +179,7 @@ outputs:
         - ld_impl_{{ target_platform }} >=2.36.1  # [linux]
         - libnsl  # [linux]
         - libuuid  # [linux]
+        - libxcrypt  # [linux]
```
jakirkham, line 182: Do we need to add this to `requirements/run` as well? Or should `libxcrypt` have `run_exports`?

Asking because of issues like this one: https://github.com/conda-forge/ray-packages-feedstock/pull/150#issuecomment-1960928574

cc @h-vetinari

jakirkham: That's not the question though

The question is why we are still needing to work around this in packages:

* https://github.com/conda-forge/ray-packages-feedstock/pull/150#issuecomment-1960928574
* https://github.com/conda-forge/torch-scatter-feedstock/pull/25#discussion_r1523989702

mbargull: > That's not the question though

That was _literally_ the question you asked.

> The question is why we are still needing to work around this in packages:

If you have downstream builds that do not find headers/libraries then the first thing to check is whether those builds respect the `CFLAGS`/`CXXFLAGS`/`LDFLAGS` we provide.

jakirkham: Hey Marcel, I'm sorry if I've said something that offended you. That was not my intent 😞

Agree that is a good thing to check 👍

Am mentioning this as we are actually seeing a few cases where users were able to build Python extensions ok (so presumably they got Python headers without issues before). However when it comes to `crypt.h` or `-lcrypt` they are running into issues:

* https://github.com/conda-forge/torchmd-net-feedstock/pull/21#issuecomment-1956007199
* https://github.com/conda-forge/libtiledb-sql-py-feedstock/pull/17#issuecomment-1954224399

While it certainly could be flag issues, I'm curious how they are able to find other things like `Python.h`, but not `crypt.h`. Would be interested in hearing your insights

isuruf: > While it certainly could be flag issues, I'm curious how they are able to find other things like Python.h, but not crypt.h. Would be interested in hearing your insights

Can you check whether Python.h and crypt.h are in the same location? Then you should have your answer.

mbargull: > Hey Marcel, I'm sorry if I've said something that offended you. That was not my intent 😞

Oh, don't worry no offense taken.
(In case my use of "literally" came over as harsh: I meant its literal ( :laughing: ) sense -- not in the often encountered hyperbolical misuse!)

The issue is just that such things grab the attention of some(/multiple) of our most valued members (maintainers of `python` package/core members) and as such tap into important/scarce resources of ours -- which I'd like us not to do if it can be avoided by doing some quick checks beforehand (which I'd kind of expect from seasoned members like yourself).

----
As Isuru hinted at, `Python.h` is not located directly under the main `include` directory; as such builds which only include paths to `Python.h`'s directory won't find other headers.
(The reason that only 3.8 builds are affected is just that later versions don't include `config.h` in `Python.h`.)
Mind you, it is actually a good thing that these builds failed then; it gives you an indication that other flags (e.g., CPU arch) that we'd want to use for our packages are left out, too.

jakirkham: Ok fair enough

My point is more that we are seeing recurring issues with a security fix where maintainers are frequently throwing up their hands and linking to issue ( https://github.com/conda-forge/linux-sysroot-feedstock/issues/52 ) with various hacks they had to implement. So am concerned something isn't work right (am not claiming to know what it is yet and it may not just be one thing), but the nature of these problems does merit more attention

> (The reason that only 3.8 builds are affected is just that later versions don't include config.h in Python.h.)
Mind you, it is actually a good thing that these builds failed then; it gives you an indication that other flags (e.g., CPU arch) that we'd want to use for our packages are left out, too.

Ah ok! This is the sort of insight I was hoping for. Thanks Marcel! 🙏

PR #676, opened 2024-02-15
[3.7] Build against libxcrypt on Linux

AaronOpfer:
Creating this PR as was suggested in discussion in https://github.com/conda-forge/linux-sysroot-feedstock/issues/52, modeled after other PRs along the same vein from @mbargull .

refs:

https://github.com/conda-forge/python-feedstock/pull/664
https://github.com/conda-forge/linux-sysroot-feedstock/issues/52

Checklist


jakirkham:
Thanks Aaron! 🙏

Would suggest adding the text below to `conda-forge.yml`'s `azure` section and re-render

https://github.com/conda-forge/python-feedstock/blob/b850a59080e59f5e82401dcd1ad9ad165581ecf9/conda-forge.yml#L11

```yaml
azure:
  settings_win:
    pool:
      vmImage: vs2017-win2016
```

AaronOpfer:
Gave that a try, but seems like no luck. @jakirkham

.azure-pipelines/azure-pipelines-win.yml
```diff
@@ -8,106 +8,45 @@ jobs:
     vmImage: vs2017-win2016
```
jakirkham, line 8: Had suggested `vs2017-win2016` as that is what we used before (please see above)

That said, it looks like [Azure deprecated this image]( https://devblogs.microsoft.com/devops/hosted-pipelines-image-deprecation/ )

conda-forge.yml
```diff
@@ -2,3 +2,7 @@ provider: {linux_aarch64: default, linux_ppc64le: default}
 conda_forge_output_validation: true
 conda_build:
   pkg_format: '2'
+azure:
+  settings_win:
+    pool:
+      vmImage: vs2017-win2016
```
jakirkham, line 8: The next best thing would be to try the 2019 image. This is what Microsoft recommends in the blog linked above

```suggestion
      vmImage: windows-2019
```

This is also consistent with what the `3.8` builds are using

https://github.com/conda-forge/python-feedstock/blob/121811a2e0a879f1fe8eb86d7ec57ee59c0d6397/conda-forge.yml#L1-L4

jakirkham (commented):
Thanks Aaron! 🙏

Provided another potential option below. Let's see if that works better

recipe/meta.yaml
```diff
@@ -139,6 +139,7 @@ requirements:
     - ncurses  # [not win]
     - libffi  # [not win]
     - libnsl  # [linux]
+    - libxcrypt  # [linux]
```
jakirkham, line 142: @beckermr , just to confirm, is this what we want here? Is there anything else we need?

beckermr: Assuming it has a proper run export, this should be everything.

beckermr: @mbargull already converted the current python versions so you can compare to that recipe to be sure.

jakirkham: That looks like what Marcel did in PR: https://github.com/conda-forge/python-feedstock/pull/664

Though would be happy to hear feedback from him

Looks like we already cover the `import crypt` test case

https://github.com/conda-forge/python-feedstock/blob/d9b57b0d84bc6772b2ec70dd73dbdc1536c6dc1e/recipe/run_test.py#L80

mbargull: I'm not against rebuilding against `libxcrypt`.
However, rebuilding this against an updated global pinning is probably not what would @AaronOpfer and others help.
We do not rebuild packages that depend on Python 3.6/3.7 anymore, which means that if we only rebuild `conda-forge::python=3.6,3.7`, we'll likely add updated dependency constraints which would be incompatible with Python extension modules on `conda-forge` that have been built against the same packages but only in older versions.
(Looking at the changes to `.ci_support/*.yaml` here, it doesn't look like much has changed, apart from the dropped `openssl=1.1.1` builds; but there could still be things like tightened `run_exports` in some of the dependencies -- i.e., someone should look at the actual constraints generated.)

Meaning, we might not want a full rerendering here to try to build against older libs (and have someone verify this by comparing with previous `python=3.6,3.7` builds).

So, to me `libcrypt`/`libxcrypt` is actually not the most important thing we should look out for here, but rather the other dependencies.

(FYI, Aaron's issue does probably not concern the actual linked `libcrypt` at all, but rather the fact that `Python.h` includes `crypt.h` for Python <=3.8; in >=3.9 it is only used for the `crypt` module and in >=3.13 it will not be used at all.)

jakirkham: That's a good point Marcel!

We can revert the bulk of the variant changes (though some things like the Visual Studio version are out of our hands)

Also if you haven't already, please read over this discussion for more context: https://github.com/conda-forge/linux-sysroot-feedstock/issues/52#issuecomment-1946615469

AaronOpfer: One more note: if CF maintainers would like to take over and directly edit this PR instead of puppeteering this conda-feedstock novice, please feel free to do so 🙂

jakirkham: > I have consulted with my users and discovered they are still using the openssl 1.1.1 variant of Python 3.7.12

Um I hope not. OpenSSL 1.1.1 has already reached [EOL]( https://www.openssl.org/blog/blog/2023/09/11/eol-111/ )

> Do we believe we would fix this by modifying the recipe/conda_build_config.yaml and rerendering as the .ci_support/README suggests? Or would we hand-edit the files as we're doing artifact archeology on an old recipe?

Was thinking we could just

```
git checkout 3.7 -- .ci_support
git add .ci_support
git commit -m 'Restore `.ci_support` from `3.7`'
```

AaronOpfer:
OK; all builds are green. The remaining question is: how do we inspect the resulting packages runtime data to make sure they're matching or reasonably close to matching the original Python 3.7.12 builds?

recipe/meta.yaml
h-vetinari: > Um I hope not. OpenSSL 1.1.1 has already reached [EOL](https://www.openssl.org/blog/blog/2023/09/11/eol-111/)

Python 3.7 never officially gained support for OpenSSL 3. We only [did](https://github.com/conda-forge/python-feedstock/issues/516#issuecomment-942220263) that in conda-forge.

.ci_support/linux_64_openssl3.yaml
```diff
@@ -15,33 +15,17 @@ cxx_compiler:
...
-  readline:
-    max_pin: x
-  sqlite:
-    max_pin: x
-  tk:
-    max_pin: x.x
-  xz:
-    max_pin: x.x
-  zlib:
-    max_pin: x.x
```
h-vetinari, line 44: Wouldn't we also need to match the pins for previous `pin_run_as_build` packages to "period-appropriate" versions in the local CBC?

AaronOpfer:
FYI, my company no longer needs this urgently. We ended up hacking up references to the old `sysroot_linux-64` for py36 and py37 targets since I tried to get the ball rolling a few weeks ago.

Issue #658, opened 2023-12-12
Building Python 3.13 dev releases + no GIL


jakirkham:
Python 3.13 has 2 alpha releases and a 3rd one due out in a week. Here's [the release schedule]( https://peps.python.org/pep-0719/ ) for more details

Of particular note, [the `--disable-gil` build flag]( https://docs.python.org/3.13/using/configure.html#cmdoption-disable-gil ) was added, which should allow building a Python with the GIL disabled. It's important to note that this mode is [not ABI compatible with the mode that keeps the GIL]( https://peps.python.org/pep-0703/#python-build-modes ). So any extension module built with Python would need to be rebuilt with the no GIL interpreter to be used with it. That said, in no GIL mode there is an environment variable to toggle GIL usage, [`PYTHONGIL`]( https://peps.python.org/pep-0703/#pythongil-environment-variable ). Eventually the ABI stability will be addressed, but this is still several Python releases away. So we will need to decide what to do before then

Naively would propose building both build modes

We may also want to explore building a couple common libraries (like NumPy, SciPy, scikit-learn) to try with. IIRC some libraries already have fixes included for no GIL. So those seem like a preferable starting point

There is [some interest in building out packages with Anaconda]( https://peps.python.org/pep-0703/#distribution ). So think we will want to coordinate

Would be interested to hear thoughts on this 🙂

ovo-Tim:
Hi, I'm trying to [build NoGil](https://github.com/ovo-Tim/nogil-recipe). I met some problems while building for Linux on github action.
Can you have a look? https://stackoverflow.com/questions/77895008/build-cpython-failed-on-github-action-output-tons-of-error-messages

Also, I have built successfully on my own computer. See: https://anaconda.org/ovo-tim/nogil

xhochy:
@ovo-Tim Can you make a pull request to the feedstock with your changes? Then we could integrate that here (and help you with CI)

ovo-Tim:
@xhochy Should I create a new branch, or just pull up github action?

xhochy:
Please make a pull request with a branch on your repository and keep using Azure Pipelines. We don't want to use GitHub Actions for package builds.

ovo-Tim:
@xhochy Hi, you may should create a branch. Than I wil make a pull request with this new branch. (I think I don't have permission to create a branch on this repository, I am not a git expert.)

xhochy:
[...]

PR #679, opened 2024-03-10
Python 3.13.0a4

xhochy:
Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -73,6 +73,7 @@ source:
       - patches/0020-Don-t-checksharedmods-if-cross-compiling.patch
       - patches/0021-Override-configure-LIBFFI.patch
       - patches/0022-Allow-linking-ncursesw-panelw-on-macOS-if-available.patch
+      - patches/0001-Set-PYTHONFRAMEWORKINSTALLNAMEPREFIX.patch
 
 build:
   number: {{ build_number }}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
-{% set version = "3.12.2" %}
-{% set dev = "" %}
+{% set version = "3.13.0" %}
+{% set dev = "a4" %}
 {% set dev_ = "" %}
 {% set ver2 = '.'.join(version.split('.')[0:2]) %}
 {% set ver2nd = ''.join(version.split('.')[0:2]) %}
@@ -46,7 +46,7 @@ source:
 {% else %}
   - url: https://www.python.org/ftp/python/{{ version }}/Python-{{ version }}{{ dev }}.tar.xz
     # md5 from: https://www.python.org/downloads/release/python-{{ ver3nd }}{{ dev }}/
-    md5: e7c178b97bf8f7ccd677b94d614f7b3c
+    md5: 57d4883f5191206a9f04ebd91e8a6ede
 {% endif %}
     patches:
       - patches/0001-Win32-Change-FD_SETSIZE-from-512-to-2048.patch
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -209,7 +209,6 @@ outputs:
         - set  # [win]
         - python -V
         - python3 -V            # [not win]
-        - 2to3 -h
         - pydoc -h
         - python3-config --help  # [not win]
         - set "PIP_NO_BUILD_ISOLATION=False"  # [win]
--- a/recipe/run_test.py
+++ b/recipe/run_test.py
@@ -50,7 +50,6 @@ import _ssl
 import _struct
 import _testcapi
 import array
-import audioop
 import binascii
 import bz2
 import cmath
@@ -77,10 +76,8 @@ if sys.platform != 'win32':
     if not (ppc64le or armv7l):
         import _curses
         import _curses_panel
-    import crypt
     import fcntl
     import grp
-    import nis
     import readline
     import resource
     import syslog
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -8,16 +8,16 @@ jobs:
     vmImage: ubuntu-latest
   strategy:
     matrix:
-      linux_64_build_typedebugchannel_targetsconda-forge_python_debug:
-        CONFIG: linux_64_build_typedebugchannel_targetsconda-forge_python_debug
+      linux_64_build_typedebugchannel_targetsconda-forge_python_313a4_debug:
+        CONFIG: linux_64_build_typedebugchannel_targetsconda-forge_python_313a4_debug
         UPLOAD_PACKAGES: 'True'
         DOCKER_IMAGE: quay.io/condaforge/linux-anvil-cos7-x86_64
-        SHORT_CONFIG: linux_64_build_typedebugchannel_tar_hcab9fe8502
-      linux_64_build_typereleasechannel_targetsconda-forge_main:
-        CONFIG: linux_64_build_typereleasechannel_targetsconda-forge_main
+        SHORT_CONFIG: linux_64_build_typedebugchannel_tar_h3a4cd8ab35
+      linux_64_build_typereleasechannel_targetsconda-forge_python_313a4:
+        CONFIG: linux_64_build_typereleasechannel_targetsconda-forge_python_313a4
         UPLOAD_PACKAGES: 'True'
         DOCKER_IMAGE: quay.io/condaforge/linux-anvil-cos7-x86_64
-        SHORT_CONFIG: linux_64_build_typereleasechannel_t_h0a44af7c38
+        SHORT_CONFIG: linux_64_build_typereleasechannel_t_hab5d1bc086
   timeoutInMinutes: 360
 
   steps:
--- a/.azure-pipelines/azure-pipelines-osx.yml
+++ b/.azure-pipelines/azure-pipelines-osx.yml
@@ -8,22 +8,22 @@ jobs:
     vmImage: macOS-11
   strategy:
     matrix:
-      osx_64_build_typedebugchannel_targetsconda-forge_python_debug:
-        CONFIG: osx_64_build_typedebugchannel_targetsconda-forge_python_debug
+      osx_64_build_typedebugchannel_targetsconda-forge_python_313a4_debug:
+        CONFIG: osx_64_build_typedebugchannel_targetsconda-forge_python_313a4_debug
         UPLOAD_PACKAGES: 'True'
-        SHORT_CONFIG: osx_64_build_typedebugchannel_targe_h343640355d
-      osx_64_build_typereleasechannel_targetsconda-forge_main:
-        CONFIG: osx_64_build_typereleasechannel_targetsconda-forge_main
+        SHORT_CONFIG: osx_64_build_typedebugchannel_targe_h268d298fd1
+      osx_64_build_typereleasechannel_targetsconda-forge_python_313a4:
+        CONFIG: osx_64_build_typereleasechannel_targetsconda-forge_python_313a4
         UPLOAD_PACKAGES: 'True'
-        SHORT_CONFIG: osx_64_build_typereleasechannel_tar_h217401936b
-      osx_arm64_build_typedebugchannel_targetsconda-forge_python_debug:
-        CONFIG: osx_arm64_build_typedebugchannel_targetsconda-forge_python_debug
+        SHORT_CONFIG: osx_64_build_typereleasechannel_tar_h3c1599cfdb
+      osx_arm64_build_typedebugchannel_targetsconda-forge_python_313a4_debug:
+        CONFIG: osx_arm64_build_typedebugchannel_targetsconda-forge_python_313a4_debug
         UPLOAD_PACKAGES: 'True'
-        SHORT_CONFIG: osx_arm64_build_typedebugchannel_ta_h18126b8f59
-      osx_arm64_build_typereleasechannel_targetsconda-forge_main:
-        CONFIG: osx_arm64_build_typereleasechannel_targetsconda-forge_main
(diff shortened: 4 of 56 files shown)


jakirkham:
Thanks Uwe! 🙏

Do we want to try no GIL here ( https://github.com/conda-forge/python-feedstock/issues/658 )?

Or do you want to focus on the regular build for now?

xhochy:
Step-by-step ;)

I plan to do:

1. See how this patch goes
2. Update to alpha 5
3. Add the nogil/freethreading variant
4. Make a variant that includes the JIT

jakirkham:
Sounds good. Sorry I'm letting my enthusiasm get the better of me 😅

xhochy:
You will have to need patience here. This is very low on my list at the moment.

jakirkham:
Yes of course. Have no expectations. Just enthusiasm 🙂

It's just exciting to see Python making these big steps

Sorry if I implied anything otherwise

mbargull:
> You will have to need patience here. This is very low on my list at the moment.

Thanks for starting prep work for this in any case, much appreciated!

h-vetinari:
We're about 1 month [away](https://peps.python.org/pep-0719/) from beta1 (feature freeze), and ~4 month away from rc1 (ABI freeze).

h-vetinari:
Just a short note that 3.13.0b1 got [released](https://discuss.python.org/t/python-3-13-0b1-now-available/52891) a couple hours ago - here's the [what's new](https://docs.python.org/3.13/whatsnew/3.13.html)

maresb:
Just curious, is the eventual goal here only to check which patches are necessary to turn this green, do you also hope to produce a pre-release that could be used by maintainers for testing?

mbargull:
@maresb, the idea is to hopefully produce builds for testing some time during the late beta phase (which we are in now) or release candidate phase.
For the release candidates, there is also the possibility for us to start building packages for the new version prior its final release; see https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4914 and issues linked therein.
If and when this will happen depends on the limited time of our contributor/maintainers (thanks to @xhochy for his work thus far!).

jjerphan:
https://github.com/conda-forge/python-feedstock/commit/3aabb6b9b7dce1e1c7ee0e7b1caf5d08ba447612 turns the build green for `linux_64`.

jjerphan:
Thank you for those efforts, Uwe!

https://github.com/conda-forge/python-feedstock/commit/3aabb6b9b7dce1e1c7ee0e7b1caf5d08ba447612 applied after 066f4e8 turns the builds green for `linux_64` on my end but we might want to support older versions of `sysroot_*` (in which case a more complex setup is required, I believe).

h-vetinari:
> but we might want to support older versions of `sysroot_*`

No, please, let cos6 rest in well-deserved peace.

isuruf:
@conda-forge-admin, rerender

jjhelmus:
If it helps, I have been building the 3.13 betas and rc1 for Anaconda (only linux-64 and osx-arm64 right now). The current recipe is in the [`py313rc1` branch of the AnacondaRecipes fork](https://github.com/AnacondaRecipes/python-feedstock/tree/py313rc1). Packages are in the `ad-testing/label/py313` channel or anaconda.org. 

Python 3.13 has an [experimental free-threading option](https://docs.python.org/3.13/whatsnew/3.13.html#free-threaded-cpython) that changes the ABI and platform tag (cp313t vs cp313) as well as the directory where 3rd party packags (e.g. site-packages) resides on POSIX (linux, macOS) platforms. The recipe above creates builds both with the GIL and with the experimental free-threading option. It would be great to align on how to approach these changes.

isuruf:
@conda-forge-admin, rerender

jjhelmus:
A few hints from building 3.13:
* The default configuration expects the system to provide libmpdec (mpdecimal). To use the bundled copy use `--with-system-libmpdec=no`. Longer term a libmpdec conda package should be created and linked.

* The python library directory for the free-threading variant includes a `t` suffix, `lib/python3.13t`.  In general this suffix is added in various locations (the shared library, directories, etc).  One option it to include the `t` in `$VERABI`  in the free-threading variant and replace various instances of `$VER` with `$VERABI`
```
if [[ ${PY_FREETHREADING} == true ]]; then
  THREAD=t
else
  THREAD=
fi

VERABI=${VER}${DBG}${THREAD}
...

```

* Because of the above, the site-packages directory is in the wrong location for `noarch: python` packages, resulting in them not being importable after install (see conda/conda#14053). A symlink to the location without a suffix in the free-threading variant is a work around:
```
if [[ ${PY_FREETHREADING} == true ]]; then
    ln -s "${PREFIX}/lib/python${VERABI}" "${PREFIX}/lib/python${VER}"
fi
```

recipe/sitecustomize.py
```diff
@@ -0,0 +1,15 @@
...
+    dirs_to_add.append(os.path.join(sys.prefix, 'lib', f'python3.{sys.version_info[1]}', 'site-packages'))
+# Workaround for https://github.com/conda/conda/issues/10969
+dirs_to_add.append(os.path.join(sys.prefix, 'lib', f'python3.1', 'site-packages'))
+# A python version independent directory that ABI3 and noarch packages can use.
+# This is unused at the moment, but keeping it here for experimentation.
+dirs_to_add.append(os.path.join(sys.prefix, 'lib', f'python', 'site-packages'))
+
+for d in dirs_to_add:
+    if os.path.exists(d):
+        site.addsitedir(d)
```
isuruf, line 15: @jjhelmus, thoughts on this?

recipe/sitecustomize.py
```diff
@@ -0,0 +1,15 @@
+import site, sys, os
+
+dirs_to_add = []
+# Workaround for https://github.com/conda/conda/issues/14053
```
rgommers, line 4: It would be good to expand this comment. I raised a question about the implications of this workaround at https://github.com/conda/conda/issues/14053#issuecomment-2280798542

recipe/sitecustomize.py
jjhelmus: I think these paths could also be added to sys.path using [`.pth` files](https://docs.python.org/3/library/site.html) which would leave `sitecustomize.py` available for other uses and would not break if `sitecustomize.py` was clobbered or deleted.

[Pex](https://pypi.org/project/pex/) loads modules from multiple directories by adding entries to sys.path. I've had experience with pex and most projects work fine with this setup. There are a few cases where the setup can fail where the packages work fine with a single site-packages directory:
* [Namespace packages](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) that use different namespace conventions.  
* Packages that load data, config or other files from other packages by navigating the filesystem. For example loading a config file using `os.path.join(os.path.dirname(os.path.dirname(mod.__file__)), "mod_data", "config.yaml")`.  This type of access really should be using `importlib.resources`.

Personally, I see benefits and downsides to this approach. 

I can imagine there may be confusion to users and developers if they see multiple site-packages directories which contain files. Which one is the correct one? All of these directories need to be examined to determine which projects are available to Python. Using symlinks clarifies this some but leaves multiple directories.

On the other hand if conda/mamba/pixie/pip places files in the correct site-packages directory then the workaround directories do not exist with this appoarch. There is no confusion or uncertainty. With the symlink approach the symlinks are always there.

One other gotya is that conda-build will set $SP_DIR to the first directory matching `lib/python?.*` that is not a symlink, https://github.com/conda/conda-build/blob/24.7.1/conda_build/utils.py#L971-L982. With this approach a noarch: python package installed into a free-threading environment will result in the directory being set incorrectly to the the `python3.13` directory. The logic that conda-build uses to set `SP_DIR` can be made more robust to avoid this.

jjhelmus:
How about adding the compatibility `site-packages` directories only when they exist? This will leave `sys.path` less cluttered  when everything is done correctly. The downside is modules installed into those locations after the interpreter starts will not be immediately available.

isuruf:
Isn't that what is already done?

isuruf:
> There are a few cases where the setup can fail where the packages work fine with a single site-packages directory:

> [Namespace packages](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) that use different namespace conventions.
> Packages that load data, config or other files from other packages by navigating the filesystem. For example loading a config file using os.path.join(os.path.dirname(os.path.dirname(mod.__file__)), "mod_data", "config.yaml"). This type of access really should be using importlib.resources.

Both of these are build system bugs that needs to be fixed

> I can imagine there may be confusion to users and developers if they see multiple site-packages directories which contain files. Which one is the correct one? All of these directories need to be examined to determine which projects are available to Python. Using symlinks clarifies this some but leaves multiple directories.

> On the other hand if conda/mamba/pixie/pip places files in the correct site-packages directory then the workaround directories do not exist with this appoarch. There is no confusion or uncertainty. With the symlink approach the symlinks are always there.

Agreed with this. This way, it's a bug with older conda/mamba/pixi and users of newer tools wouldn't be confused at all.

> One other gotya is that conda-build will set $SP_DIR to the first directory matching lib/python?.* that is not a symlink, https://github.com/conda/conda-build/blob/24.7.1/conda_build/utils.py#L971-L982. With this approach a noarch: python package installed into a free-threading environment will result in the directory being set incorrectly to the the python3.13 directory. The logic that conda-build uses to set SP_DIR can be made more robust to avoid this.

We can fix it in conda-build. The number of users of conda-build is way fewer than conda, so it's okay IMHO to ask them to update conda-build to the newest build.

jjhelmus:
> Isn't that what is already done?

Yes, the directories are already added conditionally. Sorry I missed the conditional line in my examination.

recipe/meta.yaml
```diff
@@ -116,21 +101,16 @@ outputs:
       skip: true  # [win and int(float(vc)) < 14]
       skip_compile_pyc:
         - '*.py'            # [build_platform != target_platform]
-      string: {{ dev_ }}h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}{{ linkage_nature }}{{ debug }}_cpython  # ["conda-forge" in (channel_targets or "")]
-      string: h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}{{ linkage_nature }}{{ debug }}  # ["conda-forge" not in (channel_targets or "")]
-{% if 'conda-forge' in channel_targets %}
+      string: {{ dev_ }}h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}{{ debug }}_{{ abi_tag }}_cpython
```
isuruf, line 104: @jjhelmus, shall we do
```suggestion
      string: {{ dev_ }}h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}{{ debug }}_{{ abi_tag }}
```

jjhelmus: I'm :+1: for this change. This will require an update to `python_abi` as it has a run constraint on python having `_cpython` at the end of the build string.

jjhelmus:
Should the free-threading variant be deprioritized with a track_features or a lower build number?

recipe/meta.yaml
```diff
@@ -302,19 +273,59 @@ outputs:
...
         - popd    # [unix]
 
+  - name: cpython
+    build:
+      noarch: generic
+    requirements:
+      - python {{ version }}.*
+      - python_abi * *_{{ abi_tag }}
+
+  - name: cpython-freethreading
```
jjhelmus, line 299: How about just `freethreading` or `free-threading`? The install command would then be `conda install python freethreading`

recipe/meta.yaml
```diff
@@ -302,19 +273,59 @@ outputs:
...
+    build:
+      noarch: generic
+{% if freethreading != "yes" %}
+      skip: True
+{% endif %}
+    requirements:
+      - python {{ version }}.*
+      - python_abi * *_{{ abi_tag }}
+
+  - name: cpython-nonfreethreading
```
jjhelmus, line 309: `gil`?

recipe/run_test.py
```diff
@@ -31,7 +33,6 @@
 import _csv
 import _ctypes
 import _ctypes_test
-import _decimal
```
jjhelmus, line 34: The lack of _decimal will cause the decimal module in the standard library to use the slower Python implementation. Adding `--with-system-libmpdec=no` will restore the C implementation (not sure what the Windows equivalent is)

isuruf: Using libmpdec conda package now

isuruf:
> Should the free-threading variant be deprioritized with a track_features or a lower build number?

Both are added now

recipe/meta.yaml
h-vetinari: I'd prefer keeping the `cpython-`

isuruf: cc @conda-forge/core

isuruf: @jjhelmus what do you think? I'm partial to `cpython-freethreading`, but not fixed on it.

jjhelmus: I feel like `cpython-freethreading` is a good metapackage for experts who understand the difference between cpython and python. I'm not sure in beginners would discover this package.  Personally, I think something like `freethreading` or `python-freethreading` would be more obvious to user. That said these could be added in addition to `cpython-freethreading` not in place of the meta-package.

GunnerTH28:
> 
> Checklist
> * [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
> * [ ] Bumped the build number (if the version is unchanged)
> * [ ] Reset the build number to `0` (if the version changed)
> * [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
> * [ ] Ensured the license file is being packaged.
> 
> 
> 
> 
>

PR #686, opened 2024-04-08
python 3.11.9

h-vetinari:
Freshly [released](https://discuss.python.org/t/python-3-11-9-is-available/50502), see [release notes](https://docs.python.org/release/3.11.9/whatsnew/changelog.html#python-3-11-9). It depends essentially on building against expat 2.6 (which we hadn't unvendored until 3.11) to address some CVEs, and thus, if we merge https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5640, we need to override this here in CBC. Seeing that this PR isn't merged yet, I haven't touched this here.

Finally, taking the opportunity to add stdlib here.


h-vetinari:
@conda-forge/python, could someone PTAL here?

PR #706, opened 2024-09-10
python v3.10.15

h-vetinari:

PR #708, opened 2024-09-10
python v3.8.20

h-vetinari:


isuruf:
@conda-forge-admin, rerender

h-vetinari:
AFAICT the `downstreams:` tests cause a new artefact to be registered for upload, which then fails due to the new output validation:
```
validation results:
{
  "linux-64/cython-3.0.11-py38h6d02427_0.conda": false,                 # <---
  "linux-64/libpython-static-3.8.20-h5888daf_0_cpython.conda": true,
  "linux-64/python-3.8.20-h4a871b0_0_cpython.conda": true,
  "noarch/setuptools-73.0.1-pyhd8ed1ab_0.conda": false,                 # <---
  "noarch/cpython-gil-3.8.20-hd8ed1ab_0.conda": true,
  "noarch/cpython-3.8.20-py38hd8ed1ab_0.conda": true
}
NOTE: Any outputs marked as False are not allowed for this feedstock. See https://conda-forge.org/docs/maintainer/infrastructure/#output-validation-and-feedstock-tokens for information on how to address this error.
```

h-vetinari:
I see now this is https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/342. Could you please leave some basic xrefs next time, so other people don't waste time duplicating the analysis...

xhochy (approved):

xhochy:
Also manually check the validation outputs.

recipe/meta.yaml
```diff
@@ -180,6 +179,8 @@ outputs:
         - libnsl  # [linux]
         - libuuid  # [linux]
         - libxcrypt  # [linux]
+        - xorg-libx11               # [linux]
+        - xorg-xorgproto            # [linux]
```
isuruf, line 183: @xhochy, these are supposed to be added to `ignore_run_exports` in the python recipe to avoid x11 being pulled into every python environment.

h-vetinari: The aarch tests were failing due to missing libxcb before. Probably `xorg-libx11` can be reduced to that.

h-vetinari: Confirmed in #706 that `libxcb` itself is not enough, so we need `xorg-libx11`, but ignore its run-export; `xorg-xorgproto` has no run-export.

Issue #703, opened 2024-09-10
2024 fall releases


h-vetinari:
See [here](https://discuss.python.org/t/python-3-13-0rc2-3-12-6-3-11-10-3-10-15-3-9-20-and-3-8-20-are-now-available/63161)

* [x] #709
* [x] #704
* [x] #705
* [x] #706
* [x] #711
* [x] #708

h-vetinari:
@conda-forge/python, can someone please look at #706 - #708? The only relevant change is switching windows compilers to VS2019, as VS2017 has been removed from the runner images.

rxm7706:
Something went awry here for python 3.12.6
- https://github.com/conda-forge/python-feedstock/pull/704 
Only the OSX versions seem to have made it to the conda-forge channel 
https://anaconda.org/conda-forge/python/files?version=3.12.6

h-vetinari:
Thanks for pointing that out, it was an issue with the CI setup that caused builds not to be uploaded. That has since been fixed. I've restarted the builds now and in about an hour or two they should be online.

h-vetinari:
All done now

PR #714, opened 2024-09-19
[main] switch away from cdts

h-vetinari:
Forward-port of changes in #706 that happened after the merge of #703

don't want full `xorg-libx11` as a run-time dep, so ignore it


isuruf:
Thanks

PR #732, opened 2024-10-09
fix naming inconsistency in convenience wrapper

h-vetinari:
In contrast to what the README states, there's no `python-gil` yet; this is a left-over from some naming iterations in #679.

Not bumping the build number because this isn't a dependency of any other output here, so nothing but a leaf package changes.


h-vetinari:
Ping @conda-forge/python, this is a trivial typo fix, but cures the situation that people currently cannot install an output that's described in the feedstock readme.

isuruf:
Thanks. We should fix the blog post too.

isuruf:
We should also mark the `cpython-gil` packages as broken too.

Issue #688, opened 2024-04-24
Python build with frame pointers enabled?


sk1p:
### Comment:

For profiling Python applications, using [the Linux perf support in Python 3.12](https://docs.python.org/3/howto/perf_profiling.html), it would be very helpful to have a build of Python available which has frame pointers enabled. For some additional context: https://www.brendangregg.com/blog/2024-03-17/the-return-of-the-frame-pointers.html 

As I'm using conda most of the time to install different Python versions, it would be very nice to have Python w/ frame pointers directly available as a conda package. Is this maybe already available somewhere, or is this something one can realistically build locally? Would that just be a matter of patching `recipe/build_base.sh` to include `-fno-omit-frame-pointer` and `-mno-omit-leaf-frame-pointer` in the `CFLAGS` and running `build-locally.py`? Thanks!

jjerphan:
I also think it would be worth having a build of CPython (and more generally conda-forge) with frame pointers enabled.

> [...] is this something one can realistically build locally? Would that just be a matter of patching `recipe/build_base.sh` to include `-fno-omit-frame-pointer` and `-mno-omit-leaf-frame-pointer` in the `CFLAGS` and running `build-locally.py`?

Yes, and yes (this should be sufficient).

PR #740, opened 2024-11-05
do not omit frame pointer on Linux

matthiasdiener:
Addresses #688.

Checklist


matthiasdiener:
This is ready for review. A bit more context for not omitting the frame pointer: https://fedoraproject.org/wiki/Changes/fno-omit-frame-pointer

xhochy:
I'm generally in favor of it but asking myself whether we should set this globally in the compiler activation?

matthiasdiener:
> I'm generally in favor of it but asking myself whether we should set this globally in the compiler activation?

FWIW, I'd be in favor of this as a (longer-term?) goal. In the meantime, would it make sense to add this flag just to this feedstock (at least for Python 3.12+)?

matthiasdiener:

mdekstrand:
@xhochy I think globally enabling frame pointers for `conda-forge` would be a great idea, it's really hard to profile across an entire application without them (and AFAICT PyPI binary wheels are mixed in terms of enabling or disabling frame pointers).

PR #744, opened 2024-11-26
Install python_bin in SRC_DIR

xhochy:
Checklist

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -11,62 +11,62 @@ jobs:
       linux_64_build_typedebugchannel_targetsconda-forge_python_debugfreethreadingno:
         CONFIG: linux_64_build_typedebugchannel_targetsconda-forge_python_debugfreethreadingno
         UPLOAD_PACKAGES: 'True'
-        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-cos7-x86_64
+        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-x86_64:alma9
         SHORT_CONFIG: linux_64_build_typedebugchannel_tar_h10d856d4b5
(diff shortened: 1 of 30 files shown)


xhochy:
Looks like the move to alma9 breaks the yum requirements:

```
+ /usr/bin/sudo -n yum install -y libX11 libxcb libxau
AlmaLinux 9 - AppStream                         9.0 MB/s | 8.6 MB     00:00    
AlmaLinux 9 - BaseOS                            3.3 MB/s | 3.4 MB     00:01    
AlmaLinux 9 - Extras                             19 kB/s |  13 kB     00:00    
No match for argument: libxau
  * Maybe you meant: libXau
Error: Unable to find a match: libxau
```

xhochy:
To avoid looking into CI logs: It fails in a downstream test for setuptools which wants to install `python=3.9` instead of the one built in this recipe.

beckermr:
Yes the issue is that a `downstream` test might be fore a `noarch: python` package which by default is tested only on `python_min`. There are two fixes, depending on what is more important.

1. Constrain the `downstream` tests in the `upstream` package to only run on `python_min`.
2. Remove the `python_min` requirement from the `downstream` package's test requirements.

More or less, you prefer 1 when testing the `downstream` package on `python_min` is the most important thing. You prefer 2 when testing all python versions of the `upstream` package with the `downstream` package is the most important thing. 

In this case, I suspect 2 is the right choice.

beckermr:
I am going to add this text to our docs for posterity.

beckermr:
xref: https://github.com/conda-forge/conda-forge.github.io/pull/2387

xhochy:
Thank you! Fix is then here: https://github.com/conda-forge/setuptools-feedstock/pull/371

isuruf:
@conda-forge-admin, rerender

PR #776, opened 2025-03-28
[3.10] Unvendor expat

schribl:
Checklist

Backport the patch from @isuruf to 3.10 (currently only available on 3.11+). This is relevant to also get new expat versions for e.g. security reasons without depending on Python to update/release.


schribl:

schribl:
This will require #778 to be merged (or another fix for the same problem).

If there is anything I can help with just let me know as this is also addressing a security issue by updating the currently vendored expat to a version without a known security issue.

schribl:
I rebased this now on #778 to show it builds and works with the fixes.

schribl:
Is there any way I can help with this? As this is also addressing a security issue by updating the currently vendored expat to a version without a known security issue.

recipe/patches/0026-Unvendor-expat.patch
```diff
@@ -0,0 +1,204 @@
+From 694404bc76fae025c40894c6d727ec6b70a95f84 Mon Sep 17 00:00:00 2001
+From: Thomas Bininda <schribl@gmail.com>
```
isuruf, line 2: Can you keep my credits here?
You can add yourself as a `co-authored-by` if you changed the patch.

schribl: Did update it, sorry I did not want to take any credit I just did the same changes as in the patch and created a patch with git as the existing patch for other versions did not apply cleanly.

I checked the other versions and there is no co-author so I also did not add one.

whophil:
I'm seeing some downstream issues which seem to be related to this change?

```
ImportError: $PREFIX/lib/python3.10/lib-dynload/pyexpat.cpython-310-x86_64-linux-gnu.so: undefined symbol: 
```

isuruf:
Can you post the full error?

schribl:
> I'm seeing some downstream issues which seem to be related to this change?
> 
> The problematic `python` version is `3.10.16 habfa6aa_3_cpython`
> 
> ```
> ImportError: $PREFIX/lib/python3.10/lib-dynload/pyexpat.cpython-310-x86_64-linux-gnu.so: undefined symbol: 
> ```

As mentioned it would be helpful to also paste the part what symbol is missing (I suspect it should be some `XML_*` one). Also in addition it would be nice if you could share the output from `ldd $PREFIX/lib/python3.10/lib-dynload/pyexpat.cpython-310-x86_64-linux-gnu.so` with `$PREFIX` being your conda prefix. Also run it from within the activated conda environment.

I had a look at the 3.10.16 packages with build number 1 and 3 in comparison and this changes dynamically links libexpat instead of statically as before so e.g. with a libexpat version 2.4.9 in the path I got a similar error message.

whophil:
Thanks @isuruf and @schribl. It seems my issue may not be a problem of this feedstock change after all.

The issue I see is due to some other package, https://pypi.org/project/matlabengine/, which is not installed by conda. Importing `matlab.engine` from this package has the behavior of modifying `sys.path` to include Matlab library search paths, which also happen to include `libexpat` of a very old version (1.9.0).

After the import of `matlab.engine`, attempting to do anything with `xmlrpc` in Python fails with the missing symbol error:
```
Python Error: ImportError: $PREFIX/lib/python3.10/lib-dynload/pyexpat.cpython-310-x86_64-linux-gnu.so: undefined symbol: XML_SetReparseDeferralEnabled
```

I presume this is due to the Python module finding the wrong libexpat. And that this wasn't an issue before because previously `libexpat` was statically linked.

PR #778, opened 2025-03-31
[3.10] Use older libffi as new is not working with 3.10 on windows

schribl:
Checklist

While working on #776 I discovered that the windows build for 3.10 (and 3.9) does not work on windows anymore.

I did not really find the root cause, but I checked old build logs from CI and the only major diff I could find was the libffi version. So I tried downgrading the libffi version and this apparently fixes the build.

Open question is:
Should the Linux build also downgrade or should be different versions of libffi be used per OS?

Also the other question would be what is the proper fix. I could not yet identify the exact PR in the Python repo that would fix this.

### DIFF
--- a/.ci_support/linux_64_.yaml
+++ b/.ci_support/linux_64_.yaml
@@ -20,8 +20,6 @@ cxx_compiler_version:
 - '13'
 docker_image:
 - quay.io/condaforge/linux-anvil-x86_64:alma9
-libffi:
-- '3.4'
 liblzma_devel:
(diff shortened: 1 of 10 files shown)


isuruf:
Thanks

schribl:
Thanks a lot @isuruf, should I backport the changes (force older libffi version) to 3.9 to fix the Windows build?

isuruf:
Yes, let's backport to 3.9

PR #831, opened 2025-11-02
Build Python 3.14 with zlib-ng on all platforms

hmaarrfk:
I really don't believe in fully ripping out zlib from conda-forge.

I think that a targeted approach is much better.

These patches are pretty close to something we could upstream.

xref: https://github.com/conda-forge/conda-forge.github.io/issues/2638

Checklist


isuruf (changes requested):
I'd rather we build zlib-ng in compat mode and not carry these patches.

hmaarrfk:
> pinning to an older Python 3.14 version.

This sounds like a very clear way for me.

Backends, and swaps, and build numbers are actually quite confusing for end users.

PR #875, opened 2026-06-11
[3.11] backport fix for stricter openssl

h-vetinari:
Backport fix for https://github.com/conda-forge/openssl-feedstock/issues/237


minrk (approved):

h-vetinari:
Thanks for the quick review @minrk! I'm merging this to unbreak things, we can still come up with another solution if people so desire, but at least that'll be from the basis of not having widespread breakage.

PR #842, opened 2026-01-15
Use sysconfigdata and `build-details.json` from shared library build

mgorny:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

Use the sysconfigdata from the shared library build rather than the static one. The main difference is that this ensures that programs link to the shared Python library rather than the static library. Other differences are mostly superficial (build directories, test invocation commands).

Also use the `build-details.json` from there. Unlike the static build version, it has all the data, including static and shared library, and extension build details.

Fixes #565

PR #885, opened 2026-07-28
Add a libpython3.dylib to support ABI3 applications

isuruf:
Note that this is not for extensions like numpy. It is for executbales like lldb.

Checklist

See https://github.com/conda-forge/lldb-feedstock/pull/110. cc @h-vetinari

### DIFF
--- a/recipe/install_shared.sh
+++ b/recipe/install_shared.sh
@@ -34,7 +34,15 @@ fi
 
 # create libpython3.dylib; linux builds only add this in release mode, do the same on osx
 if [[ "$target_platform" == osx-* && ${PY_INTERP_DEBUG} == no ]]; then
-  awk -F',' '$1 == "func" || $1 == "data" { print "_" $2 }' Doc/data/stable_abi.dat > stable_abi_exports.txt
+  # need to filter out windows-specific symbols & PyOS_CheckStack from
+  # https://github.com/python/cpython/blob/main/Doc/data/stable_abi.dat
+  awk -F',' '
+    ($1 == "func" || $1 == "data") &&
+    $4 != "on Windows" &&
+    $2 != "PyOS_CheckStack" {
+      print "_" $2
+    }
+  ' Doc/data/stable_abi.dat > stable_abi_exports.txt
 
   $CC -dynamiclib \
    -install_name @rpath/libpython3.dylib \
--- a/recipe/install_shared.sh
+++ b/recipe/install_shared.sh
@@ -31,3 +31,14 @@ fi
 if [[ ${PY_INTERP_DEBUG} == yes ]]; then
   ln -s ${PREFIX}/lib/libpython${VERABI}${SHLIB_EXT} ${PREFIX}/lib/libpython${VERABI_NO_DBG}${SHLIB_EXT}
 fi
+
+if [[ "$target_platform" == osx-* ]]; then
+  awk -F',' '$1 == "func" || $1 == "data" { print "_" $2 }' Doc/data/stable_abi.dat > stable_abi_exports.txt
+
+  $CC -dynamiclib \
+   -install_name @rpath/libpython3.dylib \
+   -compatibility_version 3.0 -current_version ${PY_VER}.0 \
+   -Wl,-reexport_library,${PREFIX}/lib/libpython${VERABI}.dylib \
+   -Wl,-exported_symbols_list,stable_abi_exports.txt \
+   -o ${PREFIX}/lib/libpython3.dylib
+fi
--- a/recipe/install_shared.sh
+++ b/recipe/install_shared.sh
@@ -32,7 +32,8 @@ if [[ ${PY_INTERP_DEBUG} == yes ]]; then
   ln -s ${PREFIX}/lib/libpython${VERABI}${SHLIB_EXT} ${PREFIX}/lib/libpython${VERABI_NO_DBG}${SHLIB_EXT}
 fi
 
-if [[ "$target_platform" == osx-* ]]; then
+# create libpython3.dylib; linux builds only add this in release mode, do the same on osx
+if [[ "$target_platform" == osx-* && ${PY_INTERP_DEBUG} == no ]]; then
   awk -F',' '$1 == "func" || $1 == "data" { print "_" $2 }' Doc/data/stable_abi.dat > stable_abi_exports.txt
 
   $CC -dynamiclib \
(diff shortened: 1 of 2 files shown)


h-vetinari:
Thanks, this LGTM! Once this is approved/merged, we're gonna need this on all the maintenance branches too

isuruf:
We need to figure out how to port it to dev given https://github.com/conda-forge/python-feedstock/pull/842/

h-vetinari:
> We need to figure out how to port it to dev given #842

I was mainly concerned with needing this for `python_min` and then the intervening versions (as `python_min` climbs over time); 3.15 has more time, as far as I'm concerned. 🙃

recipe/meta.yaml
```diff
@@ -247,6 +246,7 @@ outputs:
         - for f in ${CONDA_PREFIX}/lib/python*/_sysconfig*.py; do echo "Checking $f:"; if [[ `rg @[^@]*@ $f` ]]; then echo "FAILED ON $f"; cat $f; exit 1; fi; done  # [linux64 or osx]
         - test ! -f ${PREFIX}/lib/libpython${PKG_VERSION%.*}.a  # [unix]
         - test ! -f ${PREFIX}/lib/libpython${PKG_VERSION%.*}.nolto.a  # [unix]
+        - test -f ${PREFIX}/lib/libpython3${SHLIB_EXT}          # [unix]
```
h-vetinari: I think we have to restrict this to the non-debug ABI?
```suggestion
        - test -f ${PREFIX}/lib/libpython3${SHLIB_EXT}          # [unix and (build_type != "debug")]
```

isuruf:
> I was mainly concerned with needing this for python_min and then the intervening versions (as python_min climbs over time); 3.15 has more time, as far as I'm concerned. 🙃

You need `libpython3.so` at runtime of lldb, so it needs to be present for 3.15 too. However since there's a new `libpython` with 3.15, just installing `python` is not enough for a lldb compiled with 3.11

h-vetinari:
> You need `libpython3.so` at runtime of lldb, so it needs to be present for 3.15 too.

Of course. I just meant that 3.15 only goes GA in October.

isuruf:
>  I just meant that 3.15 only goes GA in October.

Yes, but adding a feature to refactor in October seems bad. We should port `libpython` output to older python versions too (but make it a dependency of `python` unlike 3.15)

h-vetinari:
I'm not saying we should leave it in a broken state; just that we'd have 3 months to fix it, whereas for the other python versions it would be useful right away

isuruf:
I really don't want to put `libpython3.dylib` in `python` and in three months move it into `libpython`. Can you send a PR to add `libpython` to `main`?

h-vetinari:
> Can you send a PR to add `libpython` to `main`?

Here's https://github.com/conda-forge/python-feedstock/pull/894

h-vetinari:
Thanks for merging #894. Would you like to rebase this PR yourself, or should I do it? My attempt in #902 looks as follows: https://github.com/conda-forge/python-feedstock/pull/902/commits/dae4337c6df167995a4990976920fc097557bcba

It wasn't enough to get CI to pass though, so I imagine this will take a bit more work.

recipe/install_shared.sh
```diff
@@ -31,3 +31,14 @@ fi
 if [[ ${PY_INTERP_DEBUG} == yes ]]; then
   ln -s ${PREFIX}/lib/libpython${VERABI}${SHLIB_EXT} ${PREFIX}/lib/libpython${VERABI_NO_DBG}${SHLIB_EXT}
 fi
+
+if [[ "$target_platform" == osx-* ]]; then
```
h-vetinari: I had
```suggestion
# create libpython3.dylib; linux builds only add this in release mode, do the same on osx
if [[ "$target_platform" == osx-* && ${PY_INTERP_DEBUG} == no ]]; then
```
because it seemed easiest to match what CPython's build system is doing on linux already (where it doesn't produce a `libpython3.so` for the debug ABI, at least not with our current build scripts). So if we want to have a debug build of `libpython3.{so,dylib}`, we'll also have to modify/patch the linux build to do that.

recipe/install_shared.sh
```diff
@@ -31,3 +31,15 @@ fi
 if [[ ${PY_INTERP_DEBUG} == yes ]]; then
   ln -s ${PREFIX}/lib/libpython${VERABI}${SHLIB_EXT} ${PREFIX}/lib/libpython${VERABI_NO_DBG}${SHLIB_EXT}
 fi
+
+# create libpython3.dylib; linux builds only add this in release mode, do the same on osx
+if [[ "$target_platform" == osx-* && ${PY_INTERP_DEBUG} == no ]]; then
+  awk -F',' '$1 == "func" || $1 == "data" { print "_" $2 }' Doc/data/stable_abi.dat > stable_abi_exports.txt
```
h-vetinari: On osx we get
```
Undefined symbols for architecture arm64:
  "_PyErr_SetExcFromWindowsErr", referenced from:
     -exported_symbol[s_list] command line option
  "_PyErr_SetExcFromWindowsErrWithFilename", referenced from:
     -exported_symbol[s_list] command line option
  "_PyErr_SetExcFromWindowsErrWithFilenameObject", referenced from:
     -exported_symbol[s_list] command line option
  "_PyErr_SetExcFromWindowsErrWithFilenameObjects", referenced from:
     -exported_symbol[s_list] command line option
  "_PyErr_SetFromWindowsErr", referenced from:
     -exported_symbol[s_list] command line option
  "_PyErr_SetFromWindowsErrWithFilename", referenced from:
     -exported_symbol[s_list] command line option
  "_PyExc_WindowsError", referenced from:
     -exported_symbol[s_list] command line option
  "_PyOS_CheckStack", referenced from:
     -exported_symbol[s_list] command line option
  "_PyUnicode_AsMBCSString", referenced from:
     -exported_symbol[s_list] command line option
  "_PyUnicode_DecodeCodePageStateful", referenced from:
     -exported_symbol[s_list] command line option
  "_PyUnicode_DecodeMBCS", referenced from:
     -exported_symbol[s_list] command line option
  "_PyUnicode_DecodeMBCSStateful", referenced from:
     -exported_symbol[s_list] command line option
  "_PyUnicode_EncodeCodePage", referenced from:
     -exported_symbol[s_list] command line option
ld: symbol(s) not found for architecture arm64
```

I think the problem is that `stable_abi.dat` contains windows-only symbols that are then not present on osx.

h-vetinari:
@isuruf, I found the fix, which isn't too complicated/intrusive IMO. Could you PTAL?

isuruf:
Thanks

h-vetinari:
FWIW, I just figured out that `libpython3.so` is not useable for this purpose, because it's essentially completely empty: https://github.com/python/cpython/issues/104612. There's a so-far unmerged [PR](https://github.com/python/cpython/pull/144136) that fixes this with a `patchelf`-based approach. We could backport that commit. Or figure out an equivalent to the macOS post-install approach.

PR #894, opened 2026-08-09
[main] Split shared Python library to `libpython`

h-vetinari:
Partial backport of #842, specifically 6eeda71c4e440d6d17a1936f522cb374d2368a27, as suggested in #885. I also picked up another useful commit from the dev branch while rebasing. Once this PR (and a rebased #885) are merged, I'm happy to backport the pair as combined PRs to the maintenance branches.

PR #907, opened 2026-08-27
actually move python*.dll to libpython on windows

h-vetinari:
Follow-up to #894. I noticed by chance that the `python3.dll` etc. did not actually end up in [`libpython`](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fwin-64%2Flibpython-3.14.7-h4f90d01_101_cp314.conda&with_broken=true).

That's despite tests to this effect
https://github.com/conda-forge/python-feedstock/blob/663c2af350b70a5d5771ff9788b6318dba71e2fd/recipe/meta.yaml#L372-L378
which however got satisfied spuriously by an additional test dependency
https://github.com/conda-forge/python-feedstock/blob/663c2af350b70a5d5771ff9788b6318dba71e2fd/recipe/meta.yaml#L353-L355

That dependency is a left-over from the `libpython` tests prior to the build order inversion in 636474b517b8f7ffe7dd4f7af0f2fa84497c0cd0. Now those tests make no sense on the `libpython` output anymore (causing us to shoehorn the full-fledged `python` back in to find e.g. `Python.h` and the executables, which also hides problems with the file-presence tests), so move them to `python`.

### DIFF
--- a/recipe/install_base.bat
+++ b/recipe/install_base.bat
@@ -51,7 +51,7 @@ if "%PY_FREETHREADING%" == "yes" (
 )
 
 :: Populate the root package directory
-for %%x in (python%VERNODOTS%%THREAD%%_D%.dll python3%THREAD%%_D%.dll python%EXE_T%%_D%.exe pythonw%EXE_T%%_D%.exe) do (
+for %%x in (python%EXE_T%%_D%.exe pythonw%EXE_T%%_D%.exe) do (
   if exist %SRC_DIR%\PCbuild\%HOST_DIR%\%%x (
     copy /Y %SRC_DIR%\PCbuild\%HOST_DIR%\%%x %PREFIX%
   ) else (
--- a/recipe/install_shared.bat
+++ b/recipe/install_shared.bat
@@ -1,0 +1,37 @@
+setlocal EnableDelayedExpansion
+echo on
+
+if "%target_platform%"=="win-64" (
+   set HOST_DIR=amd64
+)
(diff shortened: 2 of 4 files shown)


h-vetinari:
@mgorny, could I ask you to review here? With 636474b517b8f7ffe7dd4f7af0f2fa84497c0cd0, I don't think the tests for `libpython` make sense anymore as they were originally written, but I wanted to make sure I didn't misunderstand your original intent here.

mgorny (approved):
I'm pretty sure there was no special reason for that. Just a mistake made while splitting them.

isuruf:
We should revert this PR. libpython is empty in dev branch and it should be empty because python.exe depends on it.

h-vetinari:
> We should revert this PR. libpython is empty in dev branch and it should be empty because python.exe depends on it.

It's reverted in #908. For 3.14 and below, `python` depends on `libpython`, so `python.exe` will always find the required DLLs.

isuruf:
> or 3.14 and below, python depends on libpython, so python.exe will always find the required DLLs.

Yes, but it's going to be odd that libpython had the DLL for previous versions and it doesn't for 3.15+

h-vetinari:
If you want to revert this also for 3.14, then fine by me. I was just following your wishes, as well as I could interpret your comments. And you merged this PR, which I think was well-documented in terms of what it does...

isuruf:
Not blaming you here. I just forgot this detail when reviewing this PR.

PR #909, opened 2026-08-31
[3.14] Revert 907

isuruf:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #911, opened 2026-08-31
[3.14] don't move `python.dll` into `libpython` on windows

h-vetinari:
Partially reverts #907. Alternative to #909


h-vetinari:
#909 now matches this, essentially. So closing

PR #902, opened 2026-08-20
[3.13] libpython changes

h-vetinari:
backport of:
* #894
* #885
* #907
* #909

PR #913, opened 2026-09-01
[3.12] libpython changes

h-vetinari:
Same as #902, but for 3.12. Needs some modifications because e.g. no freethreading. In general, I tried redoing the commit in terms of their intent, rather than blindly cherry-picking.

Also picked up two other commits from main as they helped with cherry-picking the other commits more easily: fb8ad9b63d6453cfec15d0f85684e4c87702461c & c5f668010ed4c3acb21ca793afce5d85d1f9e575


isuruf:
Thanks

PR #914, opened 2026-09-01
[3.14] reinstate _OPTIMIZED & LTO_CFLAGS for correct flag handling in install_base.sh

h-vetinari:
Fix-up for #894 (noticed while working on #913), because the following code actually requires `_OPTIMIZED` to be set
https://github.com/conda-forge/python-feedstock/blob/ec8a1a1418cab69a3acadc3d62efffe8e24889d4/recipe/install_base.sh#L40-L59

### DIFF
--- a/recipe/install_base.sh
+++ b/recipe/install_base.sh
@@ -13,9 +13,12 @@ VER=${PKG_VERSION%.*}
 _buildd_static=build-static
 _buildd_shared=build-shared
 
+# For debugging builds, set this to no to disable profile-guided optimization
 if [[ ${PY_INTERP_DEBUG} == yes ]]; then
+  _OPTIMIZED=no
   DBG=d
 else
+  _OPTIMIZED=yes
   DBG=
 fi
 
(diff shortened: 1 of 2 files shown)


h-vetinari:
https://github.com/conda-forge/python-feedstock/pull/913/changes/6d161cc7317e319b263d65d8afea255fcd24b1f8 is a glimpse (though for 3.12) of how 636474b517b8f7ffe7dd4f7af0f2fa84497c0cd0 should have looked.

h-vetinari:
@isuruf, this is an attempt at a minimal fix for the sysconfig data, which is not getting populated as before due to missing variables after splitting the build scripts (as an aside, we could add a check for undefined variables if you'd like).

However, we may want to improve the situation beyond the minimal fix. When everything was part of a single `build_base.sh` this wasn't an issue, but after the split it's IMO not very attractive to duplicate the `LTO_CFLAGS` & `_FLAGS_REPLACE` logic on both the build & install side. Perhaps we should move it to CBC?

isuruf:
I think it's fine to duplicate for now

itamarst:
I'm seeing errors compiling Python extensions because of gcc being called with a non-existent `-partition=none` argument.

It kinda seems like `-flto-partition=none` got turned into `-partition=none` in Python source code (e.g. in `lib/python3.14/_sysconfigdata__linux_x86_64-linux-gnu.py`).

Would that be fixed by this PR, or caused by this PR? This just started happening recently in CI, I reproduced locally using:

```
libpython                   3.14.7           hdc7f604_105_cp314    conda-forge
python                      3.14.7           hcd007b5_105_cp314    conda-forge
```

Looking at the PR it kinda seems like it might be caused by this but I don't understand enough bash so before I go further figured I'd ask.

isuruf:
Fixed in https://github.com/conda-forge/python-feedstock/pull/919

PR #915, opened 2026-09-01
[3.13] reinstate _OPTIMIZED & LTO_CFLAGS for correct flag handling in install_base.sh

h-vetinari:
Fix-up for #902 (same as #914), because the following code actually requires `_OPTIMIZED` to be set
https://github.com/conda-forge/python-feedstock/blob/28c6e49db015c60825d18000cdae3f4facd8d899/recipe/install_base.sh#L40-L59

### DIFF
--- a/recipe/install_base.sh
+++ b/recipe/install_base.sh
@@ -13,9 +13,12 @@ VER=${PKG_VERSION%.*}
 _buildd_static=build-static
 _buildd_shared=build-shared
 
+# For debugging builds, set this to no to disable profile-guided optimization
 if [[ ${PY_INTERP_DEBUG} == yes ]]; then
+  _OPTIMIZED=no
   DBG=d
 else
+  _OPTIMIZED=yes
   DBG=
 fi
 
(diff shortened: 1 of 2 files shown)


yhlee-tw:
@isuruf

I think this introduced a regression in the Linux 3.13 build.

With `3.13.15-hf47f18c_102_cp313`, `sysconfig` contains the invalid `-partition=none` flag, while the other LTO-related flags are still present:

```console
$ python -c 'import sysconfig; print(sysconfig.get_config_var("CFLAGS"))'
-fno-strict-overflow -Wsign-compare -DNDEBUG -O2 -Wall -fPIC -O2 -fno-merge-constants -isystem /opt/conda/envs/test/include -fdebug-prefix-map=/home/conda/feedstock_root/build_artifacts/python-split_1788361500653/work=/usr/local/src/conda/python-split-3.13.15 -fuse-linker-plugin -ffat-lto-objects -partition=none -fPIC -O2 -fno-merge-constants -isystem /opt/conda/envs/test/include -fdebug-prefix-map=/home/conda/feedstock_root/build_artifacts/python-split_1788361500653/work=/usr/local/src/conda/python-split-3.13.15 -fuse-linker-plugin -ffat-lto-objects -partition=none
```

and pip builds of C extensions fail with:

```text
gcc: error: unrecognized command-line option '-partition=none'
```

Downgrading to `3.13.15-hf47f18c_101_cp313` fixes the issue; that build still has `-O3` and the valid `-flto-partition=none`.

It looks like the intent here is to rewrite `-O3` to `-O2` and remove the build-time LTO flags:

```text
-fuse-linker-plugin
-ffat-lto-objects
-flto-partition=none
-flto
```

In `_102`, the `-O3` -> `-O2` rewrite happened, but the LTO cleanup seems to have only partially happened: `-fuse-linker-plugin` and `-ffat-lto-objects` are still there, and `-flto-partition=none` became the invalid `-partition=none`.

EDIT: now I see the same reported in https://github.com/conda-forge/python-feedstock/pull/914#issuecomment-5514167090 (on 3.14) as well as #919 and #920.

